PR #1, opened 2016-11-29
Initial authorization setup

stackptr:


models/User.js
```diff
@@ -24,6 +24,20 @@ var userSchema = new mongoose.Schema({
 
 });
 
+
+
+userSchema.methods.verifyPassword = function(candidatePassword, cb){
```
stackptr: Schema methods are exposed when we create an instance of a schema (e.g., `var user = new User();`) and the `this` keyword refers to that current instance, with all the attributes on it. Used here because we'll always be verifying a potential password with a user we've plucked from the database

package.json
```diff
@@ -7,6 +7,7 @@
   },
   "dependencies": {
     "async": "^1.5.0",
+    "bcrypt": "^0.8.7",
```
stackptr: Important! When adding dependencies via `npm`, include the `--save` parameter so that the dependency is listed in this `package.json` file.

Dependencies are hosted by `npm` rather than synced via github. They are retrieved via `npm install` which you should use immediately following any checkout/pull/clone.

router/auth/passport.js
```diff
@@ -10,6 +13,25 @@ passport.deserializeUser(function(id, done) {
   });
 });
 
+passport.use(new LocalStrategy({
```
stackptr: Passport is very powerful middleware, and it makes sense to leverage LocalStrategy (each type of authentication method is called a Strategy, e.g. FacebookStrategy and OAuthStrategy). Passport takes care a lot of stuff in the background -- here we tell it what fields correspond to a user/password pair in the request, and then provide a callback as the actual logic to determine when a user has properly authenticated

router/auth/passport.js
```diff
@@ -10,6 +13,25 @@ passport.deserializeUser(function(id, done) {
...
+  usernameField: 'email',
+  passwordField: 'password'
+}, function(email, password, done) {
+    User.findOne({ email: email }, function (err, user) {
+      if (err) {
+        return done(err);
+      }
+      if (!user){
+        return done(null, false);
```
stackptr: If no user has been found from the given email, we pass false to indicate authentication has failed. Later in the routing, this will lead to a 401 which we can handle on the frontend)

router/auth/passport.js
```diff
@@ -10,6 +13,25 @@ passport.deserializeUser(function(id, done) {
...
+      }
+
+      user.verifyPassword(password, function(err, user){
```
stackptr: Clean abstraction of code -- passport does not need to actually understand what password verification entails. It just does some basic stuff to find a user, asks it if the passwords match, and provides a callback to handle the result

(Extra credit: there is some functional programming concepts that can be leveraged here to make this code cleaner and less redundant. See if you can find it!)

package.json
justinksheen: Hi Corey, I was just wondering what we are using the "package.json" file for. Is it simply a description of the server (similar to a README), with dependencies that are expected to be installed with "npm install"?

router/auth/passport.js
justinksheen: Is it correct to understand passport as a set of middleware functions that we pass a request to from the router?

package.json
stackptr: For our purposes yes, it defines dependencies. It also defines libraries that are distributed via npm (for example, name of the library, version number, script definitions) described here: https://docs.npmjs.com/files/package.json

Since we're not distributing this via npm, we won't be using most of those fields

router/auth/passport.js
stackptr: Exactly. Passport defines a structure for how to authenticate users that we then fill in with our custom logic, and then that acts as middleware where authentication needs to happen (some routes might not need authed users -- then simply leave out the middleware)

PR #2, opened 2017-02-24
Onboarding

justinksheen:


router/auth/index.js
```diff
@@ -44,18 +45,157 @@ module.exports = function(app){
...
+        day = req.body.day;
+
 
     if (!email || !password){
       return res.json({
         err: 'Must supply an email and password for registration'
       });
     }
 
+    if (!firstname || !lastname) {
```
stackptr: For registration purposes we don't want to require profile fields. The goal is to get them registered as simply as we can by requiring only the minimal amount of information, in this case the email and password. During onboarding we can get the profile data.

router/auth/index.js
```diff
@@ -44,18 +45,157 @@ module.exports = function(app){
...
+      });
+    }
+    if (isNaN(year) || isNaN(month) || isNaN(day)) {
+      return res.json({
+        err: 'Please use valid birth date (e.g. 01/01/1970)'
+      });
+    }
+
+    // Verify password for registration
+    if (password.length < 8) {
```
stackptr: Password verification looks great. We'd probably move this to the User model with a `verifyPasswordRequirements` method or something, which would return the error message which would then be passed as a response here. Since we're not using these password requirements anywhere else currently, this fine for right now. In the future we might want to allow a user to change their password, which means a new endpoint with the same logic, and at that point the change I'm talking about is necessary to keep code DRY

router/api/user.js
```diff
@@ -20,8 +20,16 @@ module.exports = function(router){
 		UserCtrl.update({
 			userId: req.user._id,
 			data: {
-				name: data.name,
-				picture: data.picture
+				year: data.year,
```
stackptr: Has this endpoint been tested? I'm concerned that if I don't send any of these fields in with the request (i.e., subject is blank) then it will get destroyed in the update

router/api/user.js
```diff
@@ -45,4 +53,124 @@ module.exports = function(router){
 			}
 		});
 	});
-};
+
+	router.get('/user/:picture', function(req, res){
```
stackptr: To query profile data we'll use a similar structure as setting it. Instead of multiple endpoints, we'll use one endpoint (the already existing `GET /api/user`) to send all relevant profile information

router/api/user.js
justinksheen: For this endpoint, would it be better if we split this up so that there are several "router.put(/user)"? So like, there would be one that's like "router.put(/user.year)" "router.put(/user.name)" etc?

router/api/user.js
justinksheen: Okay, I will fix this now.

router/auth/index.js
justinksheen: Commented out the code you mentioned.

router/api/user.js
stackptr: I don't think so. The reason is that we would rather update many fields at once with one request. For instance, if we have three fields, name picture and year:

If I want to update name and picture, I send in a request with the info for those two items, and don't specify year--those fields are updated, year is kept the same.

If I only want to update one field, I send in a request with only that field, and everything else is kept the same.

If we had individual endpoints for updating, the first scenario would require two requests, which adds complication to the front-end: what happens if the first request fails? What happens if the first request succeeds but we leave the page before the 2nd is made? And so on.

router/api/user.js
```diff
@@ -42,130 +42,38 @@ module.exports = function(router){
 		});
 	});
 
-	router.get('/user/:id', function(req, res){
+	router.get('/user', function(req, res){
 		UserCtrl.get({
```
stackptr: UserCtrl.get retrieves a user based on a userId. That user will already have a picture, name, etc. A GET route is meant to retrieve data (which the original route did), a POST route is used to update that data

justinksheen: I see, but doesn't the PUT endpoint (starting on line 17) take care of this, or should we make a POST route as well?

stackptr: Sure, PUT is for editing existing data, generally: http://restcookbook.com/HTTP%20Methods/put-vs-post/

PR #4, opened 2017-06-20
Password recovery WIP

ilakumar:


router/auth/index.js
```diff
@@ -30,161 +31,276 @@ module.exports = function(app){
...
+
+  if (!token){
+    return res.json({
+      err: 'No password reset token given'
+    });
+  } else if (!email || !password){
+    return res.json({
+      err: 'Must supply an email and password for password reset'
+    });
+  } else if (!newpassword){
```
stackptr: Restating this point to track on Github: not really necessary to have this endpoint take both a `password` and `newpassword`--the client can be responsible for ensuring these two passwords match and then sending the singular password along. By exposing the endpoint to two different passwords and tasking it with checking them, the server is redoing work that the UI should already take care of. It's rather subjective, I guess, but something to think about, regarding the division of labor between server and client

router/auth/index.js
```diff
@@ -30,161 +31,276 @@ module.exports = function(app){
...
+  ResetPasswordCtrl.finishReset({
+    token: token,
+    email: email
+  }, function(err, user){
+    if (err){
+      res.json({
+        err: err
+      });
+    } else {
+      user.hashPassword(password, function(err, hash){
```
stackptr: `hashPassword` should probably be done in a Ctrl function and not in the router. This was probably already happening before you touched this file, so no worries

stackptr:
Actually, do make sure that for a new user, the registrationCode is still being set properly.

models/User.js
```diff
@@ -23,7 +23,8 @@ var userSchema = new mongoose.Schema({
     default: false
   },
   verificationToken: String,
-  registrationCode: String,
+  registrationToken: String,
```
ilakumar: I changed registrationToken back to registrationCode, I must have accidentally changed that!

router/auth/index.js
```diff
@@ -30,161 +31,276 @@ module.exports = function(app){
   var router = new express.Router();
 
   router.get('/logout', function(req, res){
-      req.logout();
-      res.json({
-        msg: 'You have been logged out'
-      });
+    req.logout();
```
ilakumar: Ok! That makes sense, I think I might have fixed the indentation on this page while I was making changes-- I will be more careful next time.

router/auth/index.js
```diff
@@ -30,161 +31,276 @@ module.exports = function(app){
...
-    if (numNumber == 0) {
-      return res.json({
-        err: 'Password must contain at least one number'
+    } else if (!data.studentCode && !data.volunteerCode){
+      res.json({
+        err: 'Invalid registation code'
       });
-    }
+    } else {
+      var user = new User();
```
ilakumar: I didn't make any changes to that-- again the issue of having fixed the indentation of the whole page.

PR #5, opened 2017-08-03
Training WIP

ilakumar:


controllers/TrainingCtrl.js
```diff
@@ -0,0 +1,111 @@
+var Question = require('../models/Question');
+var User = require('../models/User');
+var ObjectId = require('mongodb').ObjectID;
+
+// Fisher-Yates shuffle
+function shuffle(array) {
```
ilakumar: That makes sense! Thanks!

controllers/TrainingCtrl.js
```diff
@@ -0,0 +1,111 @@
...
+    array[currIndex] = array[randomIndex];
+    array[randomIndex] = tempValue;
+  }
+
+  return array;
+}
+
+module.exports = {
+  getQuestions: function(options, callback){
+    var category = JSON.parse(JSON.stringify(options.category));
```
ilakumar: Oh yup you're right

controllers/TrainingCtrl.js
```diff
@@ -0,0 +1,111 @@
...
+        // sort questions by subcategory into arrays
+        questions.map(function(question) {
+          var subcategory = question.subcategory;
+          questionsBySubcategory[subcategory].push(question);
+        });
+
+        // get x unique, random objects from n objects in arrays
+        subcategories.map(function(subcategory) {
+          var questions = questionsBySubcategory[subcategory];
+          // initialize array of values from 0 to questions.length - 1
```
ilakumar: You're right!

controllers/TrainingCtrl.js
```diff
@@ -0,0 +1,111 @@
...
+          var array = [];
+          for (var i = 0; i < questions.length; i++) {
+            array[i] = i;
+          }
+          array = shuffle(array);
+
+          // change depending on how many of each subcategory are wanted
+          for (var i = 0; i < 1; i++) {
```
ilakumar: This part takes 1 question from each subcategory and creates a quiz array-- i < 1 so that it only takes one question. If we decide we want 2 questions from each subcategory, this will be changed to i < 2 etc.

controllers/TrainingCtrl.js
```diff
@@ -0,0 +1,111 @@
...
+            var index = array[i];
+            var question = questions[index];
+            randomQuestions.push(question);
+          }
+        });
+
+        randomQuestions = shuffle(randomQuestions);
```
ilakumar: Yes!

PR #6, opened 2017-08-10
Training WIP

ilakumar:


models/User.js
```diff
@@ -54,7 +54,7 @@ var userSchema = new mongoose.Schema({
   advancedCoursesText: String,
   extracurricularActivitesText: String,
 
-
+  score: Number,
```
stackptr: If there are multiple quizzes, is this one property enough?

controllers/TrainingCtrl.js
```diff
@@ -52,10 +52,8 @@ module.exports = {
           var questions = questionsBySubcategory[subcategory];
           questions = shuffle(questions);
           // change depending on how many of each subcategory are wanted
-          for (var i = 0; i < 1; i++) {
-            var question = questions[i];
-            randomQuestions.push(question);
-          }
+          const NUM_QUESTIONS = 1;
```
stackptr: Define constants at the top of the file

controllers/TrainingCtrl.js
```diff
@@ -67,34 +65,46 @@ module.exports = {
...
           var correctAnswer = question.correctAnswer;
+          idCorrectAnswerMap[question._id] = question.correctAnswer;
           var userAnswer = idAnswerMap[question._id];
           if (correctAnswer == userAnswer) {
             score = score + 1;
           }
         });
+        var hasPassed = false;
+        var percent = score / questions.length;
+        if (percent >= 0.80) {
```
stackptr: Probably a good idea to define this another constant, such as `PASS_THRESHOLD` or something

router/api/training.js
```diff
@@ -14,13 +14,14 @@ module.exports = function(router){
...
-		TrainingCtrl.getQuizScore({userid: req.body.userid, idAnswerMap: req.body.idAnswerMap}, function(err, score){
+		TrainingCtrl.getQuizScore({userid: req.body.userid, idAnswerMap: req.body.idAnswerMap, category: req.body.category}, function(err, data){
 			if (err){
 				res.json({err: err});
 			} else {
 				res.json({
           msg: 'Score calculated and saved',
-          score: score
+          score: data.score,
+					idCorrectAnswerMap: data.idCorrectAnswerMap
```
stackptr: make sure your editor is set to respect the tabs/spaces the file is using (e.g., in Atom this is Settings > Editor > Tab Type: `auto`)

ilakumar:
How should I go about switching from a JSON file to a YAML file?

controllers/TrainingCtrl.js
```diff
@@ -0,0 +1,104 @@
...
+    var idAnswerMap = options.idAnswerMap;
+    var score = 0;
+    var answer;
+    var obj_ids = Object.keys(idAnswerMap);
+    Question.find({'_id': {$in: obj_ids}}, function(err, questions) {
+      if (err){
+        return callback(err);
+      }
+      else {
+        questions.forEach(function(question) {
```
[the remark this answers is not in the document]
ilakumar: Now that there are additional methods in the forEach loop, I think it makes sense to keep this as a forEach instead of reduce.

stackptr:
There are a few different libraries for Node.js which support the parsing of YAML files. I would search around for something that is mature and well-documented

PR #7, opened 2017-10-20
Schedule

ilakumar:
Server side of interactive calendar for users to access and edit their schedule.

### DIFF
--- a/controllers/CalendarCtrl.js
+++ b/controllers/CalendarCtrl.js
@@ -1,0 +1,99 @@
+var User = require('../models/User');
+
+module.exports = {
+  initAvailability: function(options, callback){
+    var userid = options.userid;
+    var availability = {
+      Sunday: { '12a': false, '1a': false, '2a': false, '3a': false, '4a': false,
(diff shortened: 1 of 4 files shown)


controllers/CalendarCtrl.js
```diff
@@ -0,0 +1,94 @@
+var User = require('../models/User');
+
+module.exports = {
+  initAvailability: function(options, callback){
+    var userid = options.userid;
+    var availability = {
+      Sunday: { '12a': false, '1a': false, '2a': false, '3a': false, '4a': false,
```
stackptr: Would've been a good use for a for loop!

```
const times = [ '12a', '1a', '2a', ..., '11p']
const days = ['Sunday', ... 'Saturday']
const availability = days.map( day => ({
  [day]: times.map( time => ({ [time]: false })
})
)
```

Basically, map over the list of day names, and for each day then map over the list of all possible times. In both maps, we can put the mapped value as a key in an object by `[]`

So the following:
```
const myValue = 'x'`
const myObj = { [myValue]: false }
```

... will result in:

```
console.log(myObject) // {x: false}
```

PR #11, opened 2018-01-10
Set endedAt on session when disconnecting

stackptr:


stackptr:
Unclear if anything needs to happen on the client side with this change, since the client will still see that the session has disconnected.

PR #12, opened 2018-02-01
Twilio

WAftring:


models/Session.js
```diff
@@ -83,9 +83,9 @@ sessionSchema.methods.leaveUser = function(user, cb){
   // should set this.endedAt to Date.now and end the session, both users see the session ended regardless of who ended it 
   // student can receive a message telling them they can request help again 
   if (user.isVolunteer){
-    this.volunteer = null;
+    this.volunteer = user;
```
stackptr: This change has been made in #11 and shouldn't be involved in implementing Twilio

services/twilioSend.js
```diff
@@ -0,0 +1,67 @@
+		// Twilio Credentials
```
stackptr: What's with the indentation in this file? Also the braces are formatted very confusingly

services/twilioSend.js
```diff
@@ -0,0 +1,67 @@
...
+			hour = `${hour}a`
+		}
+		var days = ['Sunday', 'Monday', 'Tuesday','Wednesday','Thursday','Friday','Saturday']
+		var type = 'Math';
+		var time = `${hour-12}-${hour -11}`;
+		var avail = `availability.${days[day]}.${hour}`;
+
+		
+		//Connection variables
+		var connection = mongoose.connect(config.database);
```
stackptr: Why is a connection needed here? The DB connection should be happening in `main.js`

services/twilioSend.js
```diff
@@ -0,0 +1,67 @@
...
+			hour++;
+		}
+		if(hour > 12){
+			hour = `${hour - 12}p`
+		}
+		else {
+			hour = `${hour}a`
+		}
+		var days = ['Sunday', 'Monday', 'Tuesday','Wednesday','Thursday','Friday','Saturday']
+		var type = 'Math';
```
stackptr: Do not override `type` argument

services/twilioSend.js
```diff
@@ -0,0 +1,67 @@
...
+		//Connection variables
+		var connection = mongoose.connect(config.database);
+		var query = User.find({'serviceInterests': type,
+								[avail]: 'true'
+		})
+
+
+
+		query.exec(function (err, person) {
+			if (err) return handleError (err);
```
stackptr: Where / how is `handleError` defined?

controllers/SessionCtrl.js
```diff
@@ -1,6 +1,7 @@
 var Q = require('q');
 
 var Session = require('../models/Session');
+var twilio = require('../services/twilioSend')
```
stackptr: Never used. Also, suggest the import be named `TwilioService` to avoid confusion with `twilio` npm package.

services/twilioSend.js
stackptr: Rename file to `twilio.js`

services/twilioSend.js
```diff
@@ -0,0 +1,67 @@
...
+				var n = -1;
+				while (n < person.length-1) {
```
stackptr: This logic seems like it would be easy to run into rate throttling from Twilio's end. If we have, for example, 10 volunteers available and we try to text each one in short succession there will probably be some limiting in effect. Probably want logic to both stagger the messaging and cap it to some limit of messages

stackptr: `for (var i = 0; i++; i < person.length)` -- also probably rename 2nd arg to `persons`

services/twilioSend.js
```diff
@@ -0,0 +1,67 @@
...
+		
+		//Connection variables
+		var connection = mongoose.connect(config.database);
+		var query = User.find({'serviceInterests': type,
+								[avail]: 'true'
+		})
+
+
+
+		query.exec(function (err, person) {
```
stackptr: Avoid unnecessary `.exec`: `User.find({...}, function(err, person) { ...})`

services/twilioSend.js
```diff
@@ -0,0 +1,67 @@
...
+				};
+				})
+		
+		
+		function send(pnumb,name){
+		const accountSid = config.accountSid;
+		const authToken = config.authToken;
+
+		// require the Twilio module and create a REST client
+		const client = require('twilio')(accountSid, authToken);
```
stackptr: Do not `require` from within function export. Do `var twilio = require('twilio')` at the top and then, if necessary instantiate the `client` here: `const client = twilio(accountSid,authToken)` -- but if not necessary, do the whole thing at the top

services/twilioSend.js
```diff
@@ -0,0 +1,67 @@
...
+				while (n < person.length-1) {
+					n++;
+					var pnumb = person[n].phone;
+					var name = person[n].firstname;
+					send(pnumb,name);
+				};
+				})
+		
+		
+		function send(pnumb,name){
```
stackptr: `pnumb` => `phoneNumber` please

PR #13, opened 2018-02-01
Twilio

WAftring:


services/twilio.js
```diff
@@ -0,0 +1,64 @@
+
+var mongoose = require('mongoose')
+var config = require('../config.js')
+var User = require('../models/User')
+var twilio = require('twilio')
+const client = twilio(config.accountSid,config.authToken)
```
stackptr: more specifically `twilioClient`

services/twilio.js
```diff
@@ -0,0 +1,64 @@
...
+		//Connection variables
+		var query = User.find({'serviceInterests': type,
+								[avail]: 'true'
+		})
+
+		query.exec(function (err, persons) {
+			if (err){
+				console.log('Error conducting query: ' + err);
+			}
+			for(var i =0; i< Math.min(persons.length,3);i++){
```
stackptr: Good approach using `Math.min`, I would extract `3` out to a top-level definition, e.g. `const PERSONS_LIMIT = 3` or some other name describing what it is corresponding to

also we don't really have style guidelines in this repo but take the time to be consistent, this should read: `for (var i = 0; i < Math.min(persons.length, 3); i++) {`

services/twilio.js
```diff
@@ -0,0 +1,64 @@
...
+				var phoneNumber = persons[i].phone;
+				var name = persons[i].firstname;
+				if (phoneNumber != undefined) {
+					send(phoneNumber,name);
+				}
+				};
```
stackptr: Fix braces indentation

services/twilio.js
```diff
@@ -0,0 +1,64 @@
...
+		})
+
+		query.exec(function (err, persons) {
+			if (err){
+				console.log('Error conducting query: ' + err);
+			}
+			for(var i =0; i< Math.min(persons.length,3);i++){
+				var phoneNumber = persons[i].phone;
+				var name = persons[i].firstname;
+				if (phoneNumber != undefined) {
```
stackptr: Since people have a nullable `phone` field, it's possible that we hit the limit of 3 people without ever sending a message. We want to send the first X people a message _who have a phone number_.

Use some variable as a counter and increment every time that `send` method is called. Then in your for loop check that variable as being less than the limit variable, instead of `i`

services/twilio.js
```diff
@@ -0,0 +1,64 @@
...
+		else {
+			hour = `${hour}a`
+		}
+
+		var days = ['Sunday', 'Monday', 'Tuesday','Wednesday','Thursday','Friday','Saturday']
+		var time = `${hour-12}-${hour -11}`;
+		var avail = `availability.${days[day]}.${hour}`;
+
+
+		//Connection variables
```
stackptr: This comment is no longer valid

services/twilio.js
```diff
@@ -0,0 +1,64 @@
...
+		var phoneNumber = '+1' + phoneNumber
+		client.messages.create({
+				to: phoneNumber,
+				from: config.sendingNumber,
+				body: `Hi ${name}, A student is waiting for help in ${type} at app.upchieve.org`,
+			})
+			.then(message => console.log(`Message sent to ${phoneNumber} with message id \n` + message.sid));  		
+
+		}
+//	}
```
stackptr: Remove unnecessary commented code

WAftring:
Made requested changes

services/twilio.js
```diff
@@ -29,36 +29,36 @@ module.exports  = {
...
 				console.log('Error conducting query: ' + err);
 			}
-			for(var i =0; i< Math.min(persons.length,3);i++){
-				var phoneNumber = persons[i].phone;
-				var name = persons[i].firstname;
-				if (phoneNumber != undefined) {
+			var c = 0;
+			for (var i = 0; i < persons.length; i++) {
+				if (c === PERSON_LIMIT){
+					break;
```
stackptr: Avoid `break` -- use either `return` to terminate the `for` loop or even better make it part of the loop conditional: `... ; i < person.length && c < PERSON_LIMIT; ...`

WAftring: Change adjusted the conditional

PR #19, opened 2018-03-24
Show only certified sessions

zuoyuanh:
This branch add support for subtopics, so that front end can show volunteers the sessions they are certified to volunteer


zuoyuanh:
It should be false by default, but I’m setting it to true on purpose in order to do the testing (otherwise I will see no session on my volunteer list during my testing), and I will change it back

PR #26, opened 2018-08-05
refactored twilio

rossedwardsus:
Fixed mongo query and optimized.
Moved code into functions.
Notifications are now being sent.

### DIFF
--- a/services/twilio.js
+++ b/services/twilio.js
@@ -3,17 +3,24 @@ var mongoose = require('mongoose')
 var config = require('../config.js')
 var User = require('../models/User')
 var twilio = require('twilio')
-const twilioClient = twilio(config.accountSid,config.authToken)
+const client = twilio(config.accountSid,config.authToken)
 
-module.exports  = {
+//todo
+//limit instead of stopping at the index of 3
+//move code to separate functions
+//foreach
+//limit data response from server
+//lodash
+//ensureindex
+//logging
 
-	notify: function(type){
+function getAvailability(){
 
-		//Time check
 		var date = new Date()
 		var day = date.getDay();
 		var hour = date.getHours();
 		var min = date.getMinutes()/60
+
 
 		if(min >= .5){
 			hour++;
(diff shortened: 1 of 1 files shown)


rossedwardsus:
I pushed up a commit with the requested code.

services/twilio.js
```diff
@@ -27,37 +34,56 @@ module.exports  = {
...
+var getAvailableVolunteersFromDb = function(subtopic){
+
+		var availability = getAvailability();
+		console.log(availability);
+
+		var certification_passed = subtopic + ".passed";
 
-		var query = User.find({'serviceInterests': type,
-								[avail]: 'true'
+		var user_query = {};
```
stackptr: Prefer:

```
var userQuery = {
  certificationPassed = true,
  availability: true,
  registrationCode: "COACH18"
}
```

Or inline it directly to `User.find` since the object is not reused.

Also prefer camel casing to snake casing for consistency (`certification_passed` field should also be changed)

services/twilio.js
```diff
@@ -3,18 +3,25 @@ var mongoose = require('mongoose')
 var config = require('../config.js')
 var User = require('../models/User')
 var twilio = require('twilio')
-const twilioClient = twilio(config.accountSid,config.authToken)
+const client = twilio(config.accountSid,config.authToken)
 
-module.exports  = {
+//todo
```
stackptr: Some of this todo seems to have been done in this PR.

services/twilio.js
```diff
@@ -27,37 +34,56 @@ module.exports  = {
...
+}
+
+function send(phoneNumber, name, subtopic){
+	var phoneNumber = '+1' + phoneNumber
+	client.messages.create({
+			to: phoneNumber,
+			from: config.sendingNumber,
+			body: `Hi ${name}, A student is waiting for help in ${subtopic} at app.upchieve.org`,
 		})
-		const PERSON_LIMIT = 3;
```
stackptr: Strongly advise throttling the number of messages sent here in a predictable way. Twilio most likely does some amount of throttling but their limit might not be very comfortable for the expenses of a nonprofit.

services/twilio.js
rossedwardsus: Yes.  I don't know if we are using todos so i left it there to see whther it should be removed or not.

services/twilio.js
rossedwardsus: I think the limit is related to another issue.  @alysamurray can better explain this.

services/twilio.js
rossedwardsus: The keys have to be dynamically created.  The only way to do that is through setting them this way that i know of.

services/twilio.js
alysamurray: I believe we are still limiting the number of messages sent to be 3 in this line:

`var query = User.find(user_query).select({phone: 1, firstname: 1}).limit(3);`

services/twilio.js
alysamurray: I'm leaving this alone bc I'm not clear on what todos we've actually done. @stackptr feel free to remove the todos that we've actually done if it's obvious to you lol

PR #28, opened 2018-08-11
School autocomplete

lualparedes:
New end point:

### POST /complete

Expects the following request body:
```json
{
  "query": "string with a partial school name"
}
```

Performs a text search and returns a max of 10 matches.

The response body looks like this if no error occurred, and only two matches were found:

```json
{
  "suggestions": ["Some High School 1", "Some High School 2"]
}
```


rossedwardsus:
How is the performance on this?

models/School.js
```diff
@@ -0,0 +1,73 @@
+const mongoose = require('mongoose');
+
+const schoolSchema = new mongoose.Schema({
```
stackptr: Is this schema from some data set you've found? There is a lot here that I have to imagine won't be used, and it would be better to prune that data from the set.

controllers/CompleteCtrl.js
```diff
@@ -0,0 +1,19 @@
...
+
+module.exports = {
+
+  getSuggestions: (query, callback) => {
+    School.find({$text: {$search: query}}).limit(10).exec((err, results) => {
+      if (err) {
+        callback(err);
+      }
+      else {
+        let suggestions = [];
```
stackptr: Try to use `const` instead of `let` -- mutation is rarely necessary and leads to more robust code. In this case, `results.map` can perform the pluck of `SCH_NAME` from each `result` and return it to `suggestions` for use in the callback:

```
const suggestions = results.map((result) => result.SCH_NAME)
callback(null, suggestions)
```

Notice how I can omit the `return` and `{ }` code block from the `map` predicate because ES6 is smart enough to know what I'm saying is "take this value from the object and return it."

Issue #34, opened 2018-09-05
New needs for the styleguide


lualparedes:
- [ ] **[JS/Node]** Error handling policy
- [ ] **[Workflow]** Implement continuous integration

jmromer:
Adding the first time to Asana, thanks!

PR #35, opened 2018-09-05
Fixed Twilio bug

lualparedes:
We had a couple of issues:
- The `subTopic` wasn't passed to `twilioService.notify()`
- Twilio's `authToken` was outdated

I fixed the issues and added an extra `.catch()` statement to the `send()` function inside `twilio.js` ─**we should add a catch statement to every chain of "thenables"**, and [a comprehensive error handling policy must be created](https://github.com/UPchieve/server/issues/34) in the near future.

## (Manual) test result in local instance

Console output:
Console output

Message received: 
Message received


lualparedes:
@alymurray It works! :blush: (at least locally, lol :smile:)

It works!

PR #38, opened 2018-09-06
Config fix

lualparedes:
I only deleted the extra config file and added it to `.gitignore` to prevent future issues.


.gitignore
```diff
@@ -1,6 +1,7 @@
 node_modules
 npm-debug.log
 config.js
+"config copy.js"
```
stackptr: This is probably unnecessary, we just need to make sure to watch what is being committed. This does not catch any of the numerous permutations of `config.js`

Issue #40, opened 2018-09-24
Chat log / messages not being saved?


alymurray:
The messages array for a help session on 10/21 is empty in the database. There should have been a bunch of messages in there as you can see from the screenshot.

![image](https://user-images.githubusercontent.com/20755642/45965306-25b0bd80-bff6-11e8-94a1-92fe34a7fbb0.png)

Empty messages array in database:
![image](https://user-images.githubusercontent.com/20755642/45965334-306b5280-bff6-11e8-942f-4c75fdf8c4ff.png)

Also the messages used to be saved - here's an example from an older session:
![image](https://user-images.githubusercontent.com/20755642/45965439-758f8480-bff6-11e8-84e8-9d1d31b5acd8.png)

lualparedes:
Fixed with #44.

PR #44, opened 2018-10-15
Fix/save message

zuoyuanh:
Fix the bug that messages are not saved in db.


alymurray:
@zuoyuanh looks like this has a conflict ? can you resolve it?

lualparedes:
Solves #40.

PR #56, opened 2019-02-02
Fix web/issue/133: Set correct answer for algebra question 12

ddmck:
Fix https://github.com/UPchieve/web/issues/133

Issue was filed on the web repository.

No tests have been run.


lualparedes:
@ddmck Kudos for being proactive! Thanks :blush:

PR #61, opened 2019-02-12
Fix/text notification fix

zuoyuanh:
Fix the bug that text notification is sent at the wrong time.


alymurray:
image
Not sure what's wrong? @zuoyuanh Got this error when I tried running your branch locally

alymurray:
Nevermind - thanks David for clarifying what to do. Tested locally and seems to be sending messages at the right time.

PR #71, opened 2019-03-05
Fix/join-session

lualparedes:
The new code verifies that the number of users in a session/room is at most 2 (1 student + 1 volunteer). This is meant to solve [#134 (web)](https://github.com/UPchieve/web/issues/134).

This PR now also prevents socket events emitted by non-participating users from having an effect. Thus, non-participants are prevented from interfering even if the redirect fails.

~The frontend that handles the `reject` event will be implemented in the next weeks. I'll add a comment with a link to the frontend PR when it's done.~ The redirect is accomplished by emitting a new `bump` event to the socket of the user who tried to join.

### Random extra notes
- I also added `nodemon` to avoid having to restart the server when developing.
- I tried to replicate the standards that we set for the frontend even though we haven't set up a linter for the backend (:muscle: but we will soon... hopefully :sweat_smile: ).

Related: https://app.asana.com/0/775116702697292/1107635562909714

David's frontend PR: https://github.com/UPchieve/web/pull/172


router/api/sockets.js
```diff
@@ -25,11 +25,24 @@ module.exports = function(app) {
...
-            io.emit("sessions", SessionCtrl.getSocketSessions());
-            io.to(session._id).emit("session-change", session);
+          }
+          else {
+            const numOfMembersInSession =
+              io.nsps['/'].adapter.rooms[data.sessionId] === undefined ?
+              1 :
+              io.nsps['/'].adapter.rooms[data.sessionId].length;
+
+            if (numOfMembersInSession <= 2) {
```
jmromer: This should be strictly less than, right?

```suggestion
            if (numOfMembersInSession < 2) {
```

router/api/sockets.js
```diff
@@ -25,11 +25,24 @@ module.exports = function(app) {
...
             console.log("Could not join session");
             io.emit("error", err);
-          } else {
-            socket.join(data.sessionId);
-            console.log("Session joined:", session._id);
-            io.emit("sessions", SessionCtrl.getSocketSessions());
-            io.to(session._id).emit("session-change", session);
+          }
+          else {
+            const numOfMembersInSession =
```
jmromer: This could be a little more succinct, e.g.:

```js
            const members = io.nsps['/'].adapter.rooms[data.sessionId] || [];
            const numOfMembersInSession = members.length || 1;
```

jmromer:
@lualparedes Nice! Thanks for moving this forward. Let's keep this PR open for a bit. I've started working on the corresponding client-side changes and will open a PR against this one so we can do some manual testing before merging.

router/api/sockets.js
jmromer: Actually, I think we should take a different approach to determining the number of users in a session. Checking the session object for the presence of both student and volunteer will be more robust. Relying on the socket will allow for some error states:

**Scenario 1**
Student joins:
```js
// user joined: 5c7d4d83e4c1f38bc421bae8

io.nsps['/'].adapter.rooms[data.sessionId]
// Room { sockets: { V5R0h0dpnDeqLpdOAAAB: true }, length: 1 }
session.volunteer._id
// none yet
session.student._id
// 5c7d4d83e4c1f38bc421bae8
SessionCtrl.getSocketSessions().length
// 1
```

Volunteer joins:
```js
// user joined: 5c7d4d83e4c1f38bc421baea

io.nsps['/'].adapter.rooms[data.sessionId]
// Room {  sockets: { V5R0h0dpnDeqLpdOAAAB: true, tPNnwIFvfMki2eeTAAAD: true },  length: 2 }
session.volunteer._id
// 5c7d4d83e4c1f38bc421baea
session.student._id
// 5c7d4d83e4c1f38bc421bae8
SessionCtrl.getSocketSessions().length
// 2
```

Another volunteer joins:
```js
// user joined: 5c7d4d83e4c1f38bc421bae9

io.nsps['/'].adapter.rooms[data.sessionId]
// Room { sockets: { V5R0h0dpnDeqLpdOAAAB: true, tPNnwIFvfMki2eeTAAAD: true, tlYP8LqEaFxZCY4EAAAE: true },  length: 3 }
session.volunteer._id
// 5c7d4d83e4c1f38bc421bae9
session.student._id
// 5c7d4d83e4c1f38bc421bae8
SessionCtrl.getSocketSessions().length
// 3
```

After a restart, student reloads:
```js
// user joined: 5c7d4d83e4c1f38bc421bae8

io.nsps['/'].adapter.rooms[data.sessionId]
// Room { sockets: { mSBpySIDu50OnuTcAAAD: true }, length: 1 }.  <--- length 1
session.volunteer._id
// 5c7d4d83e4c1f38bc421baea        [new volunteer]
session.student._id
// 5c7d4d83e4c1f38bc421bae8
SessionCtrl.getSocketSessions().length
// 1            <--- length 1
```

**Scenario 2**
Student starts a session and volunteer joins:

```js
// user joined: 5c7d4d83e4c1f38bc421baea

io.nsps['/'].adapter.rooms[data.sessionId]
// Room {   sockets: { iUu26D6S1nhGgH_tAAAE: true, q_61X_7DKYAdPjPlAAAG: true },  length: 2 }
session.volunteer._id
// 5c7d4d83e4c1f38bc421baea
session.student._id
// 5c7d4d83e4c1f38bc421bae8
SessionCtrl.getSocketSessions().length
// 2
```

Both navigate away, hard-refresh, rejoin:
```js
// user joined: 5c7d4d83e4c1f38bc421baea

io.nsps['/'].adapter.rooms[data.sessionId]
// Room {  sockets:    { vZ_drxqEBQETzReRAAAR: true,  FMidpJOIUBCQn87MAAAT: true,      XzwYGV1kTnyw6niLAAAU: true,   'd2Qn6yWlRKf7Bc-pAAAW': true },   length: 4 }
session.volunteer._id
// 5c7d4d83e4c1f38bc421baea
session.student._id
// 5c7d4d83e4c1f38bc421bae8
SessionCtrl.getSocketSessions().length
// 3. 
```

jmromer: The underlying problem seems to be that duplicate sessions end up in the SessionList component, so that after a volunteer joins a session, that updated session object is emitted to the client, but so to is a duplicate session object with no volunteer. I'm digging my way toward the root cause. This may have something to do with it:

```
GET /api/user 304 40.394 ms - -
POST /api/session/check 200 33.102 ms - 59
Joining session... 5c8008098482bea4e383d1d6
Session joined: 5c8008098482bea4e383d1d6
NOTE: _sessions is an object                                                  <---------------------
Client disconnected
User 5c7d4d83e4c1f38bc421bae8 leaving from 5c8008098482bea4e383d1d6
5c8008098482bea4e383d1d6 'socketSession.isDead() => ' true
GET /api/user 304 13.906 ms - -
POST /api/session/check 200 15.302 ms - 59
Joining session... 5c8008098482bea4e383d1d6
Session joined: 5c8008098482bea4e383d1d6
WARN: _sessions has been changed to an array.                     <---------------------
```

That's from the `list` method:

```js
SessionManager.prototype.list = function(){
  if (Array.isArray(this._sessions)) {
    console.log('ERROR: _sessions has been changed to an array:')
  } else {
    console.log('NOTE: _sessions is an object');
  }

  var sessions = this._sessions;
  return Object.keys(sessions).map(function(id){
    return sessions[id].session;
  });
};
```

Very weird.

lualparedes:
@jmromer I added the field `haveJoined` to the session object to keep track of the number of users that *have joined* the session (**not** current *connected* users).

This way, we can prevent having more than one volunteer in a session ([router/api/sockets.js#L33](https://github.com/UPchieve/server/pull/71/files#diff-509f3db2e4fdd5cf85299c2de10fae73R33)) and also prevent volunteer substitution ([models/Session.js#L94](https://github.com/UPchieve/server/pull/71/files#diff-c85e4fa0da25cc36124a84aa26ca5927R94)). 

Demo

jmromer:
I'm not sure I yet see the benefit of adding an integer `hasJoined`  -- what does that approach offer over checking for the presence of `student` and `volunteer` 🤔  Is there an error state we'd avoid by using `hasJoined`.

lualparedes:
Thanks for pointing that out, I only needed to refactor the [upstream guard](https://github.com/UPchieve/server/pull/71/commits/83ec1e5d6ea1945f53a28a18b92e36b081ff38c2#diff-c85e4fa0da25cc36124a84aa26ca5927R87) to make it work in [the socket guard](https://github.com/UPchieve/server/pull/71/commits/83ec1e5d6ea1945f53a28a18b92e36b081ff38c2#diff-509f3db2e4fdd5cf85299c2de10fae73R31) :v:.

controllers/SessionCtrl.js
```diff
@@ -237,6 +237,11 @@ module.exports = {
 
       session.joinUser(user, function (err, savedSession) {
         if (err) {
+          console.log(err)
```
jkeat: @dmnisson Should we handle this error differently, in a way more accommodating of https://github.com/UPchieve/server/pull/97 ?

router/api/sockets.js
```diff
@@ -41,12 +45,30 @@ module.exports = function (app) {
...
-          } else {
-            socket.join(data.sessionId)
-            console.log('Session joined:', session._id)
-            io.emit('sessions', SessionCtrl.getSocketSessions())
-            io.to(session._id).emit('session-change', session)
+            socket.emit('bump', err)
+            return
+          }
+
+/*
```
jkeat: @dmnisson Is this commented-out block of code important to keep around? Ideally we shouldn't have large section of dead code sitting around.

controllers/SessionCtrl.js
dmnisson: I've changed this to simply execute the callback with the error, so that `router/api/sockets.js` can capture it and send it to Sentry when #97 is merged.

router/api/sockets.js
dmnisson: With the new way of verifying that the person joining is a participant in the session, we don't need to keep this around anymore. I've removed it.

PR #89, opened 2019-05-15
[Bug] Remove cleanspeak related codes

jma26:
Description: Removed cleanspeak-related code from ModerationCtrl.js. We will be reusing the same files cleanspeak was on to integrate our own moderation logic.

Asana Task Link: https://app.asana.com/0/775116702697292/1117529889872071

### DIFF
--- a/README.md
+++ b/README.md
@@ -335,13 +335,7 @@ Expects the following request body:
 }
 ```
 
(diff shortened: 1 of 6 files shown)


jma26:
Done! Removed the code and added brief comments about it :)

PR #91, opened 2019-06-07
[Feature] Chat filter

jma26:
Task: https://www.notion.so/upchieve/6abe4da08ab045eea31efb13148448d6?v=1d1d833174ef402195a1d75586a1356d&p=4019f19b10374a729511027590267522


jma26:
Links
-----
Task: https://app.asana.com/0/775116702697292/1119101986551081

Re-implemented chat filter with our own RegExp as we are no longer using CleanSpeak API. We are currently using the same controller setup, ModerationCtrl, return a boolean. Next step would be to open up the service in the frontend and possibly include a reason for the falseness in the future.

* Add phone, curse/bad words, and email regular expression
* Update README.MD explanation of POST /moderate/message section

jkeat:
companion client side PR: https://github.com/UPchieve/web/pull/174

PR #92, opened 2019-06-07
Add validators to User model

dmnisson:
https://www.notion.so/upchieve/6abe4da08ab045eea31efb13148448d6?v=1d1d833174ef402195a1d75586a1356d&p=ec4963f6260545dab3f6cfb3c95db045

I've added a validator to the phone field that accepts numbers only in the strict format ###-###-####. This is intended to make all phone records in the database consistent in form; the client can accept phone numbers in other formats and convert them so the server accepts it.

I have also now added appropriate "required" validators to the User model for fields that are required on signup, so that the document will only save if those fields are defined and non-null.

Also, I added a check in the registration handler that prevents addition of the user to the database if they did not accept the user agreement.

Fields that are required on signup are now sent to auth/register instead of api/user, so that they can be verified before the user is saved in the database.

### DIFF
--- a/controllers/UserCtrl.js
+++ b/controllers/UserCtrl.js
@@ -62,7 +62,7 @@ module.exports = {
       return callback('No fields defined to update')
     }
 
-    User.findByIdAndUpdate(userId, update, { new: true }, function (err, user) {
+    User.findByIdAndUpdate(userId, update, { new: true, runValidators: true }, function (err, user) {
       if (err) {
         return callback(err)
       } else {
--- a/models/User.js
+++ b/models/User.js
@@ -42,11 +42,10 @@ var userSchema = new mongoose.Schema({
      type: String,
      validate: {
        validator: function(v) {
-         // see http://regexlib.com/REDetails.aspx?regexp_id=58
-         var re = /^([0-9]( |-)?)?(\(?[0-9]{3}\)?|[0-9]{3})( |-)?([0-9]{3}( |-)?[0-9]{4}|[a-zA-Z0-9]{7})$/;
+         var re = /^[0-9]{3}-[0-9]{3}-[0-9]{4}$/;
          return re.test(v);
        },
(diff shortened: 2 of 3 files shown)


dmnisson:
The Asana link was incorrect. I've edited it to point to the correct task.

dmnisson:
The Asana link was incorrect. I've edited it to point to the correct task.

alymurray:
silly question @jkeat @dmnisson but can someone confirm that the format we're requiring phone numbers in is actually the format we want ? I thought we were originally sending phone numbers to twilio in this format: ########## and I think that's what we've been telling volunteers to do too.

Also, if we've changed the format we want, have we also gone through the DB and converted all existing phone numbers to this new format?

dmnisson:
> silly question @jkeat @dmnisson but can someone confirm that the format we're requiring phone numbers in is actually the format we want ? I thought we were originally sending phone numbers to twilio in this format: ########## and I think that's what we've been telling volunteers to do too.
> 
> Also, if we've changed the format we want, have we also gone through the DB and converted all existing phone numbers to this new format?

Good catch! Twilio actually accepts numbers in the E.164 standard format for international numbers, and it looks like our own twilio.send() function simply adds the +1 prefix and doesn't try to remove the hyphens or otherwise change the string that is passed to it, hence the need for storing the phone numbers without hyphens. We probably should display the phone numbers to the users with hyphens (I personally find them more readable that way) but store them in the database without, to be consistent with our previous records. Should I create another PR for these changes?

dmnisson:
We may also need to write a script that searches the whole database and converts them to the format we want, since any volunteers who entered their phone number in a different format have probably not been getting notifications.

jkeat:
Yep, a new PR would be great! @dmnisson

Also, a script to update all stored numbers is a good idea 👍  We could add a tracked directory for database utility scripts like this. What do you think?

PR #93, opened 2019-06-11
[Fix] Terminated sessions appear in volunteer session list (WIP)

ravb2019:
Links
-----
Task: https://www.notion.so/upchieve/6abe4da08ab045eea31efb13148448d6?v=1d1d833174ef402195a1d75586a1356d&p=34f41247f39547bb8a518ac94449b4cf

Related PRs: https://github.com/UPchieve/web/pull/192 (client-side cause), https://github.com/UPchieve/server/pull/106 (fixes some edge cases related to broken connections)

Description
-----------

**Completed:**
No message is sent to user in chat room when another user ends session.

**Work In Progress:**
Volunteer can see user in list even after Student ends session. Volunteer can join such sessions and keep on waiting forever. (Will work on it in near future)

- [X] PR link has been posted in the task's comments
- [X] There are no new spelling/grammar mistakes in the UI, code, or documentation
- [X] Branch has been deployed to staging, and all edge cases have been manually tested
- [X] Task and PR have been updated to show that this is ready for review


controllers/SessionCtrl.js
```diff
@@ -114,7 +114,7 @@ SessionManager.prototype.disconnect = function (options) {
 // A dead session is a session with no users connected to it.
 //
 // Return a reference to the SocketSession instance.
-SessionManager.prototype.pruneDeadSessions = () => {
+SessionManager.prototype.pruneDeadSessions = function () {
```
dmnisson: This was actually one of the two bugs that led to this nearly identical behavior. The other one is in the client, which fails to disconnect the socket when a student clicks "End Session" when no volunteer has joined. Both changes will need to be made to fix this bug.

jkeat:
hey @dmnisson, could you take a look at this merge conflict as well?

jkeat:
TODO: investigate the issue of "FirstName has joined!" message appearing twice when a user joins the session.

jkeat:
@dmnisson I like the idea of indicating via chat when someone joins or leaves, but I'm having issues with the current implementation (see the above comment^).

Additionally, from a UX perspective, it might give users the wrong impression; a user might "leave" the session but return soon on a different device.

Lastly, we don't currently have a UI indicator for a message that comes from neither user, i.e. a platform/system message. Right now, it defaults to appearing as if the other user sent the message.

For these three reasons, I'm thinking we might be better off removing the message indications for the time being. Does that sound ok to you?

dmnisson:
> @dmnisson I like the idea of indicating via chat when someone joins or leaves, but I'm having issues with the current implementation (see the above comment^).
> 
> Additionally, from a UX perspective, it might give users the wrong impression; a user might "leave" the session but return soon on a different device.
> 
> Lastly, we don't currently have a UI indicator for a message that comes from neither user, i.e. a platform/system message. Right now, it defaults to appearing as if the other user sent the message.
> 
> For these three reasons, I'm thinking we might be better off removing the message indications for the time being. Does that sound ok to you?

I've removed the code for sending these messages. In the future we might want to have a status indicator somewhere in the session view that shows whether the other user is _currently_ "away" or "active", rather than having status messages appear in the chat history.

jkeat:
@dmnisson Good idea! 👍 Yeah, we need to differentiate between whether a volunteer has joined the session and whether they're currently active in the session. Right now it's a bit confusing because we're only using one indicator to signify both.

jkeat:
Btw, any idea why the messages for users joining/leaving were being sent twice?

PR #106, opened 2019-07-09
Autocorrect phone number in database when getter is accessed.

dmnisson:
Links
-----
Task: https://www.notion.so/upchieve/Fix-server-side-phone-number-getter-for-numbers-stored-the-wrong-way-in-the-database-f842a3c890b14750925a459b7ced95f8

Related PRs: https://github.com/UPchieve/server/pull/99 (original phone number formatting), https://github.com/UPchieve/server/pull/100 (utility script to autocorrect those numbers that can be)

Description
-----------
The formatting code in the current master branch works well to ensure that phone numbers of new users are stored correctly in the database for twilio.js, but users whose phone numbers are stored incorrectly in the database before deployment will not be able to log in once master is deployed. This PR allows users who have entered their phone numbers incorrectly to still log in by preventing the virtual getters for formatted phone numbers from throwing an error in the event that it doesn't match the regex, and by autocorrecting it to the right format for the database if it is still a valid U.S. phone number.

We may also want to prompt users when they login to update their phone numbers if theirs are not valid; the script will correct valid U.S. phone numbers that aren't formatted the way Twilio wants them, but it won't correct entries that are not valid U.S. phone numbers.

- [x] PR link has been posted in the task's comments
- [x] Task and PR have been updated to show that this is ready for review

### DIFF
--- a/models/User.js
+++ b/models/User.js
@@ -484,6 +484,8 @@ userSchema.methods.verifyPassword = function (candidatePassword, cb) {
   })
 }
 
+const PHONE_REGEX = /^\s*(?:[0-9](?: |-)?)?(?:\(?([0-9]{3})\)?|[0-9]{3})(?: |-)?(?:([0-9]{3})(?: |-)?([0-9]{4}))\s*$/
+
 // virtual type for phone number formatted for readability
 userSchema.virtual('phonePretty')
   .get(function () {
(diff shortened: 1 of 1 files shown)


models/User.js
```diff
@@ -484,23 +484,43 @@ userSchema.methods.verifyPassword = function (candidatePassword, cb) {
...
+    }
     return `${area}-${prefix}-${line}`
   })
   .set(function (v) {
     // regular expression that accepts multiple valid U. S. phone number formats
     // see http://regexlib.com/REDetails.aspx?regexp_id=58
     // modified to ignore trailing/leading whitespace and disallow alphanumeric characters
-    var re = /^\s*(?:[0-9](?: |-)?)?(?:\(?([0-9]{3})\)?|[0-9]{3})(?: |-)?(?:([0-9]{3})(?: |-)?([0-9]{4}))\s*$/
-    var [, area, prefix, line] = v.match(re) || []
+    var [, area, prefix, line] = v.match(PHONE_REGEX) || []
```
jkeat: in the destructured variable assignment syntax you use, the leading comma trick (to ignore the first array item) isn't super well known, so could you add a comment to explicitly describe what that's doing?

jkeat: @dmnisson more importantly... could you look into the error that occurs when creating a user account as a student? It seems that this block is executed even when the `phonePretty` property isn't accessed, so when a student signs up (not having a phone #), it breaks b/c it tries calling `.match()` on the `undefined` phone prop.

dmnisson: Sorry I missed this! I changed the setter to check first if a phone number is, in fact, being passed to it. The auth/register endpoint tries to set phonePretty even if the phone number is not being sent to it.

dmnisson: I've added a comment which should explain this.

> in the destructured variable assignment syntax used here, the leading comma trick (to ignore the first array item) isn't super well known, so could you add a comment to explicitly describe what that's doing?

PR #104, opened 2019-07-08
new field: pastSessions for User schema and future sessions of the user are added to this array

n-mathis:
Links
-----
Task: https://www.notion.so/upchieve/keep-track-of-past-sessions-in-user-schema-8fe63957200e40ce886d59084b3766a1

Related PRs:
https://github.com/UPchieve/server/pull/108
https://github.com/UPchieve/server/pull/109

Description
-----------
- all future users will have a field, pastSessions, that will keep track of any sessions they've been a part of
- pastSessions is an array of ObjectId's of Sessions which allows easy access to look up more information about the sessions
- does not consider already existing users (will need to run something to add this field to users that don't currently have it)

Developer self-review checklist
-------------------------------
- [x] Task's requirements have been fully addressed
- [x] PR link has been posted in the task's comments
- [x] Potentially confusing code has been explained with comments
- [x] No warnings or errors have been introduced; all known error cases have been handled
- [x] Any appropriate documentation (within the code, README.md, docs, etc) has been updated
- [x] There are no new spelling/grammar mistakes in the UI, code, or documentation
- [ ] Branch has been deployed to staging, and all edge cases have been manually tested
- [x] Task and PR have been updated to show that this is ready for review

PR #111, opened 2019-07-21
Fail safe notifications

dmnisson:
Links
-----
- Task: https://www.notion.so/upchieve/Fail-safe-alert-system-78bd7caa6ae541838daec1d47abdc65c
- Dependency PRs: https://github.com/UPchieve/server/pull/104

Description
-----------
Certain “failsafe” volunteers will now always receive notifications when students create new sessions. These volunteers will also receive notifications if a volunteer does not join within 5 minutes. Failsafe volunteers do not count towards the standard number of volunteers that are contacted.

This PR includes the basic new prototypes, schema fields, and additions to `twilio.js` needed for this task. I will add voice calling functionality soon.


controllers/SessionCtrl.js
```diff
@@ -178,11 +265,11 @@ module.exports = {
     var subTopic = options.subTopic
 
     if (!userId) {
-      cb('Cannot create a session without a user id', null)
+      return cb('Cannot create a session without a user id', null)
```
dmnisson: This was the issue that @jkeat and I discussed.

controllers/SessionCtrl.js
```diff
@@ -170,6 +173,90 @@ SessionManager.prototype.getUserBySocket = function (socket) {
...
+
+// The NewSessionTimekeeper manages timing of notifications that are
+// triggered by sessions that are created but never joined by volunteers
+var NewSessionTimekeeper = function () {
+  this._newSessionTimeouts = {} // sessionId => newSessionTimeout
+}
+
+// set a timeout for a session that can be cancelled if a volunteer joins
+NewSessionTimekeeper.prototype.setSessionTimeout = function (session, delay, cb, ...args) {
+  let timeout = setTimeout((...a) => {
```
jkeat: is this `timeout` variable modified at all within the `setSessionTimeout() { ... }` scope? if not, it's a good convention to default to using `const` over `let`.

generally, we should move toward using `const` and `let` throughout the whole server repo -- whenever you have new variable declarations, avoid using `var`. a good rule of thumb is to use `const` just as you would use `var`, and then if there's an error from a `const` variable being reassigned a value, you can change it to a `let`. though, i think that anytime you think you have to use `let`, you should think about whether there's a way to refactor the logic to avoid it and use `const` instead. keep in mind `const` is scoped and not global to the whole file. there's usually a better way to format things where you don't need to use `let` to do any variable reassignment.

none of these comments are very important to the PR's main content! but they help with maintainability and readability :) overall, the PR looks really good 🎉 i'll try to finish reviewing & testing it tomorrow.

dmnisson: Hi Jack,

Thanks for this feedback. I've searched the code and changed the `let` declarations that I committed to `const` where I didn't have to use `let`. There are still several `var` declarations throughout our repository, and refactoring them might be a task for another PR.

jkeat: Sounds good!

and yeah, agreed, we need to do a general code formatting refactor of the server at some point, similar to what Justin did for the web repo.

services/twilio.js
```diff
@@ -59,39 +62,149 @@ var getAvailableVolunteersFromDb = function (subtopic, options) {
...
+  var desperate = options.desperate
+
+  var voice = options.voice
+
+  var isTestUserRequest = options.isTestUserRequest
+
+  var testUserNotice = isTestUserRequest ? '[TEST USER] ' : ''
+
+  let messageText
+  if (desperate) {
```
jkeat: Would `desperate` ever be `false`? It seems like calling `sendFailsafe` is itself an indication of `desperate: true`. If we removed it, we wouldn't need this if/else statement that leads to a lot of hard-to-read and duplicated string building.

router/twiml/index.js
```diff
@@ -0,0 +1,27 @@
+const express = require('express')
```
jkeat: Could you add a brief description of this router section's general purpose? i.e. describe that we build a URL pointing to our own server and then send that to Twilio, who uses it when making a call, etc etc

Also it'd probably be good to include something similar in `sendVoiceMessage() { ... }`. The logistics of making the call aren't immediately obvious / readable.

services/twilio.js
dmnisson: `sendFailsafe` is called both when the student first begins a new session _and_ when the student is desperate, i.e. has been waiting. The `desperate` option is there to distinguish between these two types of calls so that a different message can be sent. If we want, we could use template literal syntax to make certain parts of the text depend on the `desperate` value, rather than having this if-else block. This would avoid repetition but I'm not sure if it would be more readable.

The reason the `sendFailsafe` function is called for both of these situations is that the Notion task seems to specify that failsafe volunteers should receive a "special" notification different from the one we send to "normal" volunteers, so I created a special function to send these messages. If we wanted the first message that failsafe volunteers get to be the exact same one that the available "non-failsafe" volunteers get, then we wouldn't need the `desperate` option or the if/else block--we could just update the text in the `send` function to whatever we want in the new message, so that `sendFailsafe` only sends "desperate" messages.

jkeat: Oh I see, thanks for the clarification. I think this setup is good for now :)

router/twiml/index.js
dmnisson: I added some comments describing how our system interacts with the Twilio voice calling functionality by sending Twilio a URL to call to obtain the TwiML it needs. Do these clarify things enough?

models/User.js
```diff
@@ -377,6 +377,16 @@ var userSchema = new mongoose.Schema({
     type: Boolean,
     default: false
   },
+  isFailsafeVolunteer: {
+    type: Boolean,
+    default: false,
+    validate: {
+      validator: function (v) {
+        return this.isVolunteer || !v
```
jkeat: 👍

jkeat:
@jkeat  TODO: Twilio configuration to allow phone call notifications

PR #116, opened 2019-08-01
Adding dev flags for text notifications

keerthanamuniraj:
Links
-----
- Task: https://www.notion.so/upchieve/Flag-text-notifications-with-the-environment-they-re-sent-from-when-the-env-isn-t-production-c21cb63886fa4500acd446bfdb93f9b7

Description
-----------
- Slightly altered the text notification sent to coaches to reflect the correct(dev/staging) environment URL the request is coming from
- This will remove the confusion of which environment the request is coming from


services/twilio.js
```diff
@@ -67,8 +67,9 @@ var getAvailableVolunteersFromDb = function (subtopic, options) {
 }
 
 function send (phoneNumber, name, subtopic, isTestUserRequest) {
-  var testUserNotice = isTestUserRequest ? '[TEST USER] ' : '';
-  var textBody = `${testUserNotice}Hi ${name}, a student just requested help in ${subtopic} at app.upchieve.org. Please log in now to help them if you can!`
+  var environmentUrl = config.NODE_ENV==='Dev' ? 'http://localhost:8080' : 'http://167.99.228.44:3000/'
```
jkeat: there should be three options here; development, staging, and production. right now, this will default to sending production volunteers a link to the staging URL.

i noticed that there's actually a `config.js` property for the client URL, so you could just say `var environmentUrl = config.client.host`. much simpler!

services/twilio.js
```diff
@@ -67,8 +67,9 @@ var getAvailableVolunteersFromDb = function (subtopic, options) {
 }
 
 function send (phoneNumber, name, subtopic, isTestUserRequest) {
-  var testUserNotice = isTestUserRequest ? '[TEST USER] ' : '';
```
jkeat: i know this is a little convoluted, but the `isTestUserRequest` boolean is separate from the idea of the environment. we use these "test users" on _production_ for testing session functionality on prod -- but since we don't want to notify the real volunteers when these test accounts make requests for help, we only send text alerts to admins: https://github.com/UPchieve/server/pull/112/files#diff-ced912a1af7cfbc380411df3571f7698R52

point is, there are two separate `[FLAG]`s in texts that we might want to add (possibly both, if it's a test user in a non-production environment):
1. if `isTestUserRequest === true`, include the `[TEST USER]` flag
2. if `config.NODE_ENV !== "production"`, include the `[DEVELOPMENT REQUEST]` flag.

PR #118, opened 2019-08-08
Highschool database

dmnisson:
Links
-----
- Task:https://www.notion.so/upchieve/During-signup-students-should-select-their-exact-high-school-from-a-database-of-all-U-S-schools-fb16b881c4284a80adefaf6abfb4c9a7
- Web repo PR: https://github.com/UPchieve/web/pull/206
- Database utility: https://github.com/UPchieve/server/pull/119

Data source: https://nces.ed.gov/ccd/pubschuniv.asp
Direct CSV ZIP link: https://nces.ed.gov/ccd/Data/zip/ccd_sch_029_1819_w_0a_04082019_csv.zip

Description
-----------
Currently, the signup process is open to any student who enters their school name—any school name—when they sign up. However, we will in the future want to restrict access to the app to students from schools that we approve. This PR builds upon a database model from an earlier work in progress that allows us to store U. S. high schools from the [NCES Common Core of Data Public School Universe Survey](https://nces.ed.gov/ccd/pubschuniv.asp) and adds functionality for storing manually entered schools, and assigning them a unique 8-digit identifier that can be used by students during the signup process to distinguish schools that have the same name. (Not all schools that are manually entered will have an identifier in the NCES database). The PR also implements business logic for searching this database and storing emails to which to send notification messages when their school is approved.

- [ ] Task's requirements have been fully addressed
- [ ] All new code complies with our ESLint standards


models/User.js
```diff
@@ -44,7 +44,13 @@ var userSchema = new mongoose.Schema({
     required: [function () { return this.isVolunteer }, 'Phone number is required.']
   },
 
-  highschool: { type: String, required: [function () { return !this.isVolunteer }, 'High school is required.'] },
+  highschoolName: String,
```
jkeat: Hey @dmnisson, do we need this field for any reason?

jkeat:
TODO:
Only send the necessary School info to the client, i.e. the school's name/district/state & its upchieveId

jkeat:
TODO:
In addition to adding a reference to the School on the student's User model, we should add the student's User model to an array on the School called something like "studentUsers"

controllers/SchoolCtrl.js
```diff
@@ -0,0 +1,37 @@
+const School = require('../models/School')
+
+module.exports = {
+  // search for schools by name or ID
+  search: function (query, cb) {
+    if (query.match(/^[0-9]{8}$/)) {
```
jkeat: If you enter a string with special characters like `foo bar (`, then this throws the error `SyntaxError: Invalid regular expression: /foo bar (/: Unterminated group` -- the query's special characters need to be escaped.

dmnisson: I've added a helper function to escape these special characters in the RegExp.

dmnisson:
> TODO:
> 
> Create a virtual prop called something like `searchableName` that combines the city name + school name. i.e. `${city} ${schoolName}`
> 
> The high school I went to is called "West High," but a ton of other high schools have that name, so it'd be good to have a special prop for the regex match which includes some more info ("Madison West High" is more searchable)

Document queries only execute on the fields defined in the schema. To search schools this way we would have to first retrieve the entire dataset and then filter it in our controller by the value of this virtual prop, which would be very inefficient. It would probably be better to put a separate field in the UI allowing the student to enter a city name to narrow the search. (We may even want to add a select menu for the state.)

jkeat:
@dmnisson Right, good point.

I think for now we should stick to what we have now. Some schools that we partner with might have default names that are hard to search for, in which case we can use the manually-entered stored name field to make it more searchable. Example: default of "District of Columbia International School" --> stored nickname of "DC International School"

PR #131, opened 2019-09-06
Fix: Notification references in Session object

dmnisson:
Links
-----
- Task: https://www.notion.so/upchieve/Stored-notifications-follow-up-7c957b5cf6384f06bc80ed20be623db1

Description
-----------
The logic for saving references to the `Notification` objects in the `Session` model involved loading the array into memory, pushing the new object into the array, and saving it to the database. This resulted in race conditions when multiple notifications were sent concurrently, because the function `Session.addNotification` was operating on an in-memory array that was not necessarily up-to-date with the array in the database. This PR does two things to address this problem:

1. Changes `Session.addNotification` to work directly with the MongoDB database to add the reference, rather than working with an in-memory JavaScript array.
2. Changes `twilio.notify` to add the notifications in one database operation after they have all been sent, rather than each one separately after it is sent.

- [ ] Task's requirements have been fully addressed
- [ ] Posted a link to this PR on the Notion task

### DIFF
--- a/services/twilio.js
+++ b/services/twilio.js
@@ -233,11 +233,11 @@ module.exports = {
           // early exit
           return
         }
-        
+
         console.log(persons.map(person => person._id))
         // notifications to record in the Session instance
         const notifications = []
-        
+
         async.each(persons, (person, cb) => {
           // record notification in database
           const notification = new Notification({
@@ -256,10 +256,10 @@ module.exports = {
           if (err) {
             console.log(err)
           }
-          
+
           // save notifications to Session instance
           session.addNotifications(notifications)
-          
+
           // failsafe notifications
           this.notifyFailsafe(student, type, subtopic, options)
         })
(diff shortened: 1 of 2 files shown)


services/twilio.js
```diff
@@ -229,33 +227,54 @@ module.exports = {
...
+        (err) => {
+          if (err) {
+            console.log(err)
+          }
+
+          // save notifications to Session instance
+          session.addNotifications(notifications)
+
+          // failsafe notifications
+          this.notifyFailsafe(student, type, subtopic, options)
```
jkeat: Nice, this is a good step toward being able to inform failsafes how many regular notifications were sent 👍 

One thing to watch out for: `Session.addNotifications` is asynchronous, so if `notifyFailsafe` is called synchronously right after this, there might be a chance for a race condition when `Session.notifications.length` is checked.

PR #139, opened 2019-09-25
perform null check before trying to call user.populate

dmnisson:
Links
-----

Description
-----------
The code in `router/auth/passport.js` that puts the `pastSessions` into the deserialized User object breaks when a volunteer tries to sign up because it tries to call `populate()` on a null `user`. This PR checks the truthiness of the user before trying to call `populate()`.

- [ ] Posted a link to this PR on the Notion task

### DIFF
--- a/router/auth/passport.js
+++ b/router/auth/passport.js
@@ -12,7 +12,7 @@ passport.deserializeUser(function (id, done) {
     if (err || !user) {
       return done(err, user)
     }
-    
+
     user.populate('pastSessions').execPopulate((err, populatedUser) => {
       const parsedUser = populatedUser.parseProfile()
       done(err, parsedUser)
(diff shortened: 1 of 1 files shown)


jkeat:
@dmnisson We discussed moving the action of populating `pastSessions` closer to the point we return the user data, i.e. `GET /user` and `/login`. This will avoid unnecessarily populating `pastSessions` for requests that don't need past session data. Does that sound good to you? If so, we can just use this branch to do that.

dmnisson:
> @dmnisson We discussed moving the action of populating `pastSessions` closer to the point we return the user data, i.e. `GET /user` and `/login`. This will avoid unnecessarily populating `pastSessions` for requests that don't need past session data. Does that sound good to you? If so, we can just use this branch to do that.

Yes, this sounds like a good idea.

jkeat:
> > @dmnisson We discussed moving the action of populating `pastSessions` closer to the point we return the user data, i.e. `GET /user` and `/login`. This will avoid unnecessarily populating `pastSessions` for requests that don't need past session data. Does that sound good to you? If so, we can just use this branch to do that.
> 
> Yes, this sounds like a good idea.

FYI I'm just going to merge this now and address the `pastSessions` population issue in a different PR

PR #156, opened 2019-11-11
Reorganization of session logic

dmnisson:
Links
-----
- Task: https://www.notion.so/upchieve/Session-Controller-Twilio-service-refactor-ecd838c592db4d4a8c074713a8f2bb13
- Web repo PR: https://github.com/UPchieve/web/pull/248

Description
-----------

* Session logic is restructured so that `SessionCtrl.js` can be concerned primarily with validating user input and deciding which services process it. Domain-specific logic and rules are delegated to the appropriate services that handle them.
* New services are created to allow socket events to be sent in response to events other than socket events received, including REST API calls. This simplifies the logic of synchronizing the state of the session between clients.
* Management of active sessions is done by querying the database rather than relying upon an in-memory store of Session objects which may not accurately reflect the state of the system.
* Requests from test student users will now query test volunteer users instead of admin users.
* Security of socket events in sessions is improved with [passport.socketio](https://www.npmjs.com/package/passport.socketio).

### DIFF
--- a/controllers/SessionCtrl.js
+++ b/controllers/SessionCtrl.js
@@ -47,6 +47,11 @@ module.exports = function (socketService) {
       if (!session) {
         throw new Error('No session found')
       }
+      
+      if (session.endedAt) {
+        // Session has already ended (the other user ended it)
+        return session
+      }
 
       this.verifySessionParticipant(session, user, new Error('Only session participants can end a session'))
 
--- a/controllers/SessionCtrl.js
+++ b/controllers/SessionCtrl.js
@@ -1,532 +1,129 @@
-var Session = require('../models/Session')
-var User = require('../models/User')
-var twilioService = require('../services/twilio')
+const Session = require('../models/Session')
+const User = require('../models/User')
 
-var config = require('../config')
+const sessionService = require('../services/SessionService')
+const twilioService = require('../services/twilio')
 
-// A socket session tracks a session with its users and sockets
-var SocketSession = function (options) {
-  this.session = options.session
-  this.users = [] // [User]
-  this.sockets = {} // userId => socket
-}
+const config = require('../config')
 
-// Add a socket and user to the session. If the user already has a socket,
-// disconnect and replace it
-SocketSession.prototype.join = function (options) {
-  var user = options.user
-  var socket = options.socket
-  var userIndex = this.users.findIndex(function (joinedUser) {
-    return joinedUser._id === user._id
-  })
-
-  let oldSocket
-  if (userIndex !== -1) {
-    oldSocket = this.sockets[user._id]
-
-    this.users.splice(userIndex, 1)
-  }
-
-  this.users.push(user)
-  this.sockets[user._id] = socket
-
-  // try to prevent disconnecting the user if the new socket is
-  // the same as the old one
-  if (oldSocket && oldSocket.id !== socket.id) {
-    oldSocket.disconnect(0)
-  }
-}
-
-SocketSession.prototype.leave = function (socket) {
-  var userId
-
-  Object.keys(this.sockets).some(function (socketUserId) {
-    if (this.sockets[socketUserId] === socket) {
-      userId = socketUserId
-      return true
-    }
-  }, this)
-
-  console.log('User', userId, 'leaving from', this.session._id)
-
-  var userIndex = this.users.findIndex(function (joinedUser) {
-    return joinedUser._id === userId
-  })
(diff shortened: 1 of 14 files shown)


router/auth/session-store.js
```diff
@@ -0,0 +1,26 @@
...
+const config = require('../../config')
+
+module.exports = function (app) {
+  const sessionStore = new MongoStore({
+    url: config.database,
+    autoReconnect: true,
+    collection: 'auth-sessions'
+  })
+
+  app.use(
```
jkeat: Just curious, did you separate this out from `router/auth/index.js` purely for organizational improvement?

services/SocketService.js
```diff
@@ -0,0 +1,114 @@
...
+      }
+    },
+
+    // update the list of sessions displayed on the volunteer web page
+    updateSessionList: async function () {
+      const sessions = await Session.getUnfulfilledSessions()
+      io.in('volunteers').emit('sessions', sessions)
+    },
+
+    emitNewSession: async function (session) {
```
jkeat: Was the thinking here that we might display a client-side notification to active volunteers?

controllers/SessionCtrl.js
```diff
@@ -1,539 +1,131 @@
...
-            session: session,
-            user: user,
-            socket: socket
-          })
+      this.verifySessionParticipant(session, data.user, new Error('Only session participants are allowed to send messages'))
 
-          if (user.isVolunteer) {
-            newSessionTimekeeper.clearSessionTimeouts(session)
-          }
+      await session.saveMessage(message)
```
jkeat: Do we need to `await` the message being saved? It'll save pretty quickly so it's not a huge concern, but if there's no reason to wait then we may as well deliver the message sooner.

router/api/sockets.js
```diff
@@ -95,180 +75,197 @@ module.exports = function (app) {
...
 
-    socket.on('drawing', function (data) {
+    socket.on('drawing', async function (data) {
       if (!data || !data.sessionId) return
-      SessionCtrl.verifySessionParticipantBySessionId(data.sessionId, data.user, function (err) {
-        if (err) return
-        socket.broadcast.to(data.sessionId).emit('draw')
-      })
+
+      await sessionCtrl.verifySessionParticipantBySessionId(
```
jkeat: Not directly related to this PR, but I feel like it's really intense to await `verifySessionParticipantBySessionId` for every single `drawing` event. It results in calling `await Session.findById(sessionId)` hundreds of times just to draw a line on the whiteboard.

I agree that we should make our socket communications secure, but there's got to be a more efficient way to do this. Thoughts?

services/twilio.js
```diff
@@ -265,166 +380,54 @@ module.exports = {
       })
   },
 
-  // notify both standard and failsafe volunteers
-  notify: function (student, type, subtopic, options, cb) {
-    const session = options.session
+  // begin notifying non-failsafe volunteers for a session
+  beginRegularNotifications: async function (session) {
```
jkeat: This is great! 👏

router/auth/session-store.js
dmnisson: `passport.socketio` needs to be provided the session store to use on initialization. This logic was split into a separate file so that `router/api/sockets.js` can pass to `passport.socketio` the same session store used to configure the session middleware in Express.

services/SocketService.js
dmnisson: Yes, I was thinking that we might want to visually notify logged-in volunteers when someone creates a new session, so I went ahead and added this socket event.

controllers/SessionCtrl.js
dmnisson: I'm not sure how that `await` got in there. It isn't really necessary, so I removed it.

services/SocketService.js
jkeat: Good idea 👍

router/auth/session-store.js
jkeat: Ahhh right, understood 👍

jkeat:
@dmnisson I should've put the "WIP" label on this after leaving comments so you could change it back to Ready for Review later.

But since I didn't -- is this ready again?

dmnisson:
> @dmnisson I should've put the "WIP" label on this after leaving comments so you could change it back to Ready for Review later.
> 
> But since I didn't -- is this ready again?

Yes, it's ready for your review again.

PR #169, opened 2019-12-11
Validate input before saving availability object

dmnisson:
Links
-----
- Issue: https://github.com/UPchieve/web/issues/272

Description
-----------
- Refactor CalendarCtrl.js to have a common area where names of days and hours are stored, so that this logic is not repeated.
- Change `updateAvailability` to validate the object sent by the client so as to prevent the availability subdocument in the database from being corrupted.


dmnisson:
This PR will not be necessary if we go with https://github.com/UPchieve/server/pull/172

jkeat:
Addressed by https://github.com/UPchieve/server/pull/172 instead, as noted above ^

PR #172, opened 2019-12-16
Change availability object to a subdocument schema

dmnisson:
Links
-----
- Issue: https://github.com/UPchieve/web/issues/272

Description
-----------
- The type of the `availability` field in the `User` schema is changed from `Mixed` to a subdocument schema so that availability objects are validated by default before they are saved to the database.
- Default values are added to the availability hours in the subdocument schemas, obviating the need for the `initAvailability` function in `CalendarCtrl.js`
- This PR renders https://github.com/UPchieve/server/pull/169 obsolete since the validation is now performed automatically by Mongoose.


models/User.js
```diff
@@ -49,6 +49,34 @@ const tallyVolunteerPoints = (volunteer) => {
...
+  '3p': { type: Boolean, default: false },
+  '4p': { type: Boolean, default: false },
+  '5p': { type: Boolean, default: false },
+  '6p': { type: Boolean, default: false },
+  '7p': { type: Boolean, default: false },
+  '8p': { type: Boolean, default: false },
+  '9p': { type: Boolean, default: false },
+  '10p': { type: Boolean, default: false },
+  '11p': { type: Boolean, default: false }
+}, { _id: false })
```
jkeat: Good idea to opt-out of `_id`s for these 👍

dmnisson:
> Hmm, when I put this on staging and created a new volunteer account, the resulting User instance was missing the `availability` property.

Good catch. I had specified defaults for the individual fields, but did not specify that the `availability` property needed to be set to a subdocument by default. This should be fixed now.

PR #173, opened 2019-12-17
Send user information to client in all session-change events

dmnisson:
Links
-----
- Issue: https://github.com/UPchieve/web/issues/283

Description
-----------
Sends the necessary information to the client in all `session-change` events, including when sessions end.

Issue https://github.com/UPchieve/web/issues/283 occurred because the session's `volunteer` was not being populated when it was sent in the `session-change` message, and therefore simply contained the ObjectId as a string; the client then tried to get the `_id` of this string and got `undefined`, which subsequent code interpreted as a session with no volunteer, causing the user to be redirected to the dashboard. Since the client was originally designed to receive the name of the other user in the `session-change` event I simply changed the code to send this information in all `session-change` events.

### DIFF
--- a/services/SocketService.js
+++ b/services/SocketService.js
@@ -3,6 +3,22 @@ const Session = require('../models/Session')
 
 const userSockets = {} // userId => socket
 
+/**
+ * Get session data to send to client for a given session ID
+ * @param sessionId
+ * @returns the session object
+ */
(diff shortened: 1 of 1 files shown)


jkeat:
Nice job @dmnisson! Looks good, just gonna test it on staging real quick

PR #177, opened 2019-12-19
Define virtual properties in Message model for client to access

dmnisson:
Links
-----
- Issue: https://github.com/UPchieve/web/issues/279
- Web repo PR: https://github.com/UPchieve/web/pull/287

Description
-----------
- Define virtual properties in the `Message` schema that can be used by the client to obtain information needed for displaying it properly, i.e. the user's name, the picture URL, and whether the user is a volunteer.
- Change `SocketService` to populate `user` field as needed for the socket events

### DIFF
--- a/controllers/SessionCtrl.js
+++ b/controllers/SessionCtrl.js
@@ -108,9 +108,9 @@ module.exports = function (socketService) {
 
       this.verifySessionParticipant(session, data.user, new Error('Only session participants are allowed to send messages'))
 
-      session.saveMessage(message)
+      const savedMessage = await session.saveMessage(message)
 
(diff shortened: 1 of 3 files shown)


dmnisson:
> Nice job! Just see one small issue: messages don't display a timestamp until you refresh the page
> 
> edit: might need to `await` the `session.saveMessage(message)` line in the session controller's `message` function?

I've now changed `SessionCtrl.js` to wait for the message to be saved to the database before delivering it. The timestamp was actually not being sent to the client even before this PR; what was actually happening was that the client was passing `undefined` to `moment` and thus displaying the time that the message was received according to the user's system clock, and not the actual time that the message was created.

PR #179, opened 2019-12-31
Enable the use of Docker.

ti2ger92:
Description
-----------
- Adds ability and instructions to launch server using Docker, removing dependancies on host systems, and allowing a faster start.


ti2ger92:
Note: all checks are passing here

PR #188, opened 2020-01-11
Use appropriate HTTP status codes

dmnisson:
Links
-----
- Web repo PR: https://github.com/UPchieve/web/pull/314

Description
-----------
- API endpoints were using the HTTP 200 status code for error conditions. The responses are now updated to include appropriate HTTP status codes.
- HTTP 500 error codes are intentionally left out so as to avoid conflicts with https://github.com/UPchieve/server/pull/187

### DIFF
--- a/router/api/session.js
+++ b/router/api/session.js
@@ -56,7 +56,7 @@ module.exports = function (router, io) {
       const session = await Session.findById(sessionId).exec()
 
       if (!session) {
-        res.json({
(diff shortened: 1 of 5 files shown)


ti2ger92:
BTW, I also learned that 204 is supposed to be the response for "success with no response payload", but that's arguably so unknown that it's not good to use in open source, up to you.

Issue #189, opened 2020-01-15
Schedule save API takes between 3 and 7 seconds


ti2ger92:
The API at the code index takes from 3 to 7 seconds or more to invoke both on local and production instances.

https://github.com/UPchieve/server/blob/3c716f73de0aaff49c646e4faaf1d424b8f49080/router/api/calendar.js#L4

Adhoc experiments seem to show some time taken with database save operations, but also a lot of time is taken before the API is invoked at all. This is far too long and creates a bad user experience.

jkeat:
Note: before reaching the `/calendar/save` route handler, the request goes through the passport middleware and fetches the user object

https://github.com/UPchieve/server/blob/6fb6df6ce59831ebc696b079ac86038777d5b7f8/router/auth/passport.js#L10

PR #207, opened 2020-01-27
Calendar speedup

dmnisson:
Links
-----
- Issue: https://github.com/UPchieve/server/issues/189

Description
-----------
- Instead of sending only the user ID to the controller, unnecessarily requiring the controller to fetch the user from the database, the user object from passport is sent to the controller so that the controller does not have to retrieve the user from the database in order to save the new availability data.
- This improves performance somewhat, but calendar save still takes 2-4 seconds on staging. This is probably due in part to the fact that two API endpoints, `/api/calendar/save` and `/api/calendar/tz/save`, are called for the single user action of clicking "Save", and they both require passport authentication. We may want to consider combining these into one API endpoint so that authentication only needs to be performed once.
- Another contributing factor to the slowdown may be the call to `user.availability.set` in `CalendarCtrl.updateAvailability`. This function ensures that empty objects do not cause data to be erased, but also appears to take a significant amount of time (about 0.5-1 second).

### DIFF
--- a/controllers/CalendarCtrl.js
+++ b/controllers/CalendarCtrl.js
@@ -1,45 +1,30 @@
-var User = require('../models/User')
-
 module.exports = {
   updateAvailability: function(options, callback) {
-    var userid = options.userid
-    var availability = options.availability
-    User.findOne({ _id: userid }, function(err, user) {
+    const user = options.user
+    const availability = options.availability
+
+    user.availability.set(availability)
+
+    user.save(function(err, user) {
       if (err) {
-        return callback(err)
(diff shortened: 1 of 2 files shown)


jkeat:
1. Having separate API calls for timezone and availability is definitely not great. I think that would be a good candidate for a separate issue/PR -- does that sound alright?

2. Is the alternative to calling `user.availability.set` to just do `user.availability = ...` and then save `user`? Might be worth doing this if it could shave off up to a second. (we'd still probably want to perform the empty object check before saving)

PR #209, opened 2020-01-30
Recent availability modified

treykris:
Description
-----------
- Updated the User model to have an `availabilityLastModifiedAt` property
- Set the `availabilityLastModifiedAt` property to the current date when a user makes changes to their availability
- Added a migration for existing users to contain the `availabilityLastModifiedAt`

### DIFF
--- a/dbutils/update-availability-modified.js
+++ b/dbutils/update-availability-modified.js
@@ -8,6 +8,7 @@ dbconnect(mongoose, function() {
       const pendingUpdatedUsers = listOfUsers.map(user => {
         if (!user.availabilityLastModifiedAt) {
           user.availabilityLastModifiedAt = user.createdAt
+          console.log(`Updating availability for user: ${user._id}`)
         }
         return user.save()
       })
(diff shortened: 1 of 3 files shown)


dbutils/update-availability-modified.js
```diff
@@ -0,0 +1,25 @@
...
+const User = require('../models/User')
+
+dbconnect(mongoose, function() {
+  User.find({ isVolunteer: true })
+    .then(listOfUsers => {
+      const pendingUpdatedUsers = listOfUsers.map(user => {
+        if (!user.availabilityLastModifiedAt) {
+          user.availabilityLastModifiedAt = user.createdAt
+        }
+        return user.save()
```
treykris: @jkeat,

Speaking of performance, we can probably skip the saving of users if they already have the `availabilityLastModifiedAt` property. I assume this util won't be ran much though, right?

jkeat: Yeah that might've been part of the problem with the certification dbutils script. Though, under the hood, Mongoose might not actually execute any database operations if there's nothing to update.

I have an idea for a quicker way to do this one -- we can discuss during our call

PR #210, opened 2020-02-03
Last activity at

treykris:
Description
-----------
- `lastActivityAt` property was added to the User schema to keep track of the last time a user was active 
- Created middleware to update the user's `lastActivityAt` property which is only updated after a day of elapsed time since their last activity
- Added a migration for upgrading and downgrading

### DIFF
--- a/router/api/index.js
+++ b/router/api/index.js
@@ -1,6 +1,6 @@
-var express = require('express')
-var passport = require('../auth/passport')
-var User = require('../../models/User.js')
+const express = require('express')
+const passport = require('../auth/passport')
+const addLastActivity = require('../../middleware/add-last-activity')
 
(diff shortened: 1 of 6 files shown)


router/api/index.js
```diff
@@ -1,12 +1,34 @@
...
+var User = require('../../models/User.js')
 
 module.exports = function(app, sessionStore) {
   console.log('API module')
 
   const io = require('./socket-server')(app)
 
   const router = new express.Router()
+
+  app.use(function(req, res, next) {
```
treykris: @jkeat,

I was thinking about creating a middleware folder and importing this function instead so that in the future we can have the ability to test some of our middleware

jkeat: Sounds good ^

PR #220, opened 2020-02-09
fixed 3 min elapsed to one day

treykris:
Description
-----------
- Fixed typo that allowed a user's `lastAcitivity` property to be updated after three minutes instead of one day


jkeat:
Good catch!

Issue #200, opened 2020-01-17
Simplify format of failsafe notifications


jkeat:
First text message to failsafe volunteers should be simplified to the following format:
```
Student: [First Name]
High School: [School Name]
Subject: [Subject]
First Ever Request: [Yes/No]
```

Second text, after 5 mins (if session hasn't been fulfilled):
```
Request by Student [First Name] still not filled.
Volunteers texted: [# volunteers texted so far]
```

jkeat:
Resolved by https://github.com/UPchieve/server/pull/223

PR #223, opened 2020-02-10
Simplify failsafe SMS notifications

dmnisson:
Links
-----
- Issue: https://github.com/UPchieve/server/issues/200

Description
-----------
- The failsafe SMS messages are simplified to reduce their length, according to the format given in https://github.com/UPchieve/server/issues/200.
- The second failsafe message is also changed to clarify that the number of _non-failsafe_ volunteers, not all volunteers, is given in the message.


jkeat:
@dmnisson We removed the `await`s from the initial notification kickoffs a while back in order to speed up the session request flow for students. As a result, the failsafe messages have been displaying the # of regular volunteers notified as 0.

I pushed a commit that I think will fix this issue without slowing down the session request process for the client. How does that look to you?

dmnisson:
> @dmnisson We removed the `await`s from the initial notification kickoffs a while back in order to speed up the session request flow for students. As a result, the failsafe messages have been displaying the # of regular volunteers notified as 0.
> 
> I pushed a commit that I think will fix this issue without slowing down the session request process for the client. How does that look to you?

This looks good to me.

PR #224, opened 2020-02-13
Viewed materials action

treykris:
Links
-----
- ~Dependency PRs: https://github.com/UPchieve/server/pull/221~
- Web repo PR: https://github.com/UPchieve/web/pull/329

Description
-----------
- Keep track of when a user views the review materials


router/api/training.js
```diff
@@ -38,4 +46,16 @@ module.exports = function(router) {
...
+    const { id } = req.user
+    const { category } = req.params
+
+    try {
+      await UserActionCtrl.viewedMaterials(id, category)
+    } catch (error) {
+      console.error(error)
+    }
+
+    res.sendStatus(204)
```
treykris: @jkeat,

I tried using `res.redirect` but the response kept stalling:
Screen Shot 2020-02-13 at 12 37 12 PM

router/api/training.js
```diff
@@ -51,4 +51,16 @@ module.exports = function(router) {
       }
     )
   })
+  router.get('/training/review/:category', async function(req, res, next) {
+    const { id } = req.user
+    const { category } = req.params
+
+    try {
+      await UserActionCtrl.viewedMaterials(id, category)
```
jkeat: Could we just fire this off without `await`ing it? Less delay on the client side would be nice

router/api/training.js
```diff
@@ -51,15 +51,13 @@ module.exports = function(router) {
...
   })
-  router.get('/training/review/:category', async function(req, res, next) {
+  router.get('/training/review/:category', function(req, res, next) {
     const { id } = req.user
     const { category } = req.params
 
-    try {
-      await UserActionCtrl.viewedMaterials(id, category)
-    } catch (error) {
+    UserActionCtrl.viewedMaterials(id, category).catch(error =>
```
treykris: @jkeat ,

Updated file -
I had completely forgot to remove the `await` and add Sentry. I removed the `try/catch` block so that it could be similar to the other usesActions with the appended `catch` at the end

PR #225, opened 2020-02-13
Tracking user agent related descriptions

treykris:
Links
-----
- Dependency PRs: https://github.com/UPchieve/server/pull/221

Description
-----------
- Keep track user device, browser, and operating system


controllers/SessionCtrl.js
```diff
@@ -85,8 +86,24 @@ module.exports = function(socketService) {
       }
 
       try {
+        const isInitialVolunteerJoin = !session.volunteer
+
         await session.joinUser(user)
 
+        if (isInitialVolunteerJoin && user.isVolunteer) {
+          await UserActionCtrl.joinedSession(user._id, session._id, userAgent)
```
jkeat: What about `const isInitialVolunteerJoin = user.isVolunteer && !session.volunteer` ?

treykris: @jkeat,

Nice catch, that reads much better!

models/Session.js
```diff
@@ -127,8 +126,7 @@ function failJoin(session, user, error) {
 
 // this method should callback with an error on attempts to join by non-participants
 // so that SessionCtrl knows to disconnect the socket
-sessionSchema.methods.joinUser = async function(user) {
-  let isInitialVolunteerJoin = false
+sessionSchema.methods.joinUser = function(user) {
```
jkeat: Good call removing the async-ness of this function, I'm just gonna remove the `await` from where it's used

PR #257, opened 2020-03-16
Update onboarding

treykris:
Links
-----
- Web repo PR: https://github.com/UPchieve/web/pull/349

Description
-----------
- Update the user model to send data related to if a user is onboarded or not

### DIFF
--- a/models/User.js
+++ b/models/User.js
@@ -229,7 +229,7 @@ var userSchema = new mongoose.Schema(
       default: availabilitySchema
     },
     timezone: String,
-    availabilityLastModifiedAt: { type: Date, default: Date.now },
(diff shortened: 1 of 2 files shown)


models/User.js
```diff
@@ -605,6 +607,21 @@ userSchema.virtual('mathCoachingOnly').get(function() {
   )
 })
 
+userSchema.virtual('isOnboarded').get(function() {
```
treykris: @alymurray 

Opted to go with computing the `isOnboarded` property instead of placing it directly on the user model. Since `isOnboarded` is only completed once a combination of steps are done, it seems to make more sense to keep it as a computed value

PR #259, opened 2020-03-18
Reset password edge cases

treykris:
Links
-----
- Web repo PR: https://github.com/UPchieve/web/pull/352

Description
-----------
- Added a route to verify the password reset token

### DIFF
--- a/router/auth/index.js
+++ b/router/auth/index.js
@@ -477,5 +477,31 @@ module.exports = function(app) {
     )
   })
 
+  router.post('/reset/verify', async (req, res, next) => {
+    const { token } = req.body
+
+    if (!token.match(/^[a-f0-9]{32}$/)) {
+      return res.status(422).json({
+        err:
+          'Please verify that this url matches the link that was sent to your inbox.'
+      })
+    }
+
+    try {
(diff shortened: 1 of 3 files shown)


router/auth/index.js
```diff
@@ -477,5 +477,31 @@ module.exports = function(app) {
     )
   })
 
+  router.post('/reset/verify', async (req, res, next) => {
+    const { token } = req.body
+
+    if (!token.match(/^[a-f0-9]{32}$/)) {
+      return res.status(422).json({
+        err:
+          'Please verify that this url matches the link that was sent to your inbox.'
```
treykris: @jkeat 

Thoughts on this error message copy? 

I was unsure if I should have the error message scoped to what's explicitly throwing the error, which would be an "invalid  password reset token" message and then let the client-side handle delivering a more user-friendly message or if it is fine to show a user-friendly message from the server as done above

jkeat: @treykris I don't think it's important to be user friendly with the error message from the server, especially if we're not displaying it on the client-side (which is good). I think it's better to just say what literally happened, instead of assuming we know why the error occurred

PR #267, opened 2020-03-26
Waiting period

treykris:
Links
-----
- Web repo PR: https://github.com/UPchieve/web/pull/361

Description
-----------
- Added a route to get the latest session for a student


router/api/session.js
```diff
@@ -109,4 +109,23 @@ module.exports = function(router, io) {
       next(err)
     }
   })
+
+  router.route('/session/latest').post(async function(req, res, next) {
+    const data = req.body || {}
+    const userId = ObjectId(data.user_id)
+
+    try {
+      const latestSession = await Session.findLatest({ student: userId })
```
jkeat: The `Session.findLatest` method populates the `student` and `volunteer` fields on the Session. 

Right now, the client is only using this endpoint to determine whether the latest session happened within the last 5 mins, so I think it'd be better to avoid the unnecessary population.

PR #268, opened 2020-03-26
delay notifications by one min if new student

treykris:
Description
-----------
- New student users must wait one minute before we text volunteers

### DIFF
--- a/services/twilio.js
+++ b/services/twilio.js
@@ -418,11 +418,7 @@ module.exports = {
     const populatedSession = await Session.findById(session._id)
       .populate('student')
       .exec()
-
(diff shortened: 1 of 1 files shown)


treykris:
@jkeat,

Updated to remove `delayTime`!

Thanks for pointing out the time issue! The comment above makes complete sense and like you said - adding the `setTimeout` to delay the interval by one minute would get pretty confusing. 

I mistakenly thought of just offsetting the amount of time that we make a new student wait on the `setInterval`, but I wasn't thinking of subsequent intervals are really just set 3 minutes apart after the initial `setInterval` fires. (duh trey, the function is called set interval!!)

Issue #265, opened 2020-03-25
Missing high school seed data


jkeat:
Need to add seed data for high schools and import it in `init.js`. At least one of the seed high schools should have `isApproved` set to true (and at least one false).

Also, need to add README instructions for full import; need to run:
`node dbutils/import-zip-codes.js`

ti2ger92:
The loaded schools allow the e2e tests to get further, but not pass. I think perhaps entries in the Zip Code collection are also needed to accomplish this.

@jkeat , perhaps you can add language about the full import util. I don't think it should be something Github community members should need to worry about, so maybe it can be in its own section? I think anything a Github community member needs to run basic tests should load automatically with one of our setup scripts.

Issue #266, opened 2020-03-25
Missing zip code seed data


jkeat:
Add seed data for zip codes and import it in `init.js`. At least one of the seed zip codes should have a very low median income, and one should have a very high median income. 

Also, add README instructions for full import; need to run:
`node dbutils/import-zip-codes.js`

jkeat:
Resolved by https://github.com/UPchieve/server/pull/274

PR #270, opened 2020-03-29
seed test high schools

ti2ger92:
Links
-----
- Issue: https://github.com/UPchieve/server/issues/265

Description
-----------
- Imports 2 schools as seed data. I used the Mongoose schema because I wanted the ability to "update if exists, and create if it does not exist". I added mongodb-extended-json to support extended JSON (required for Mongo's dates at rest). I hope that dependency is useful to import users without mongoimport as well.

### DIFF
--- a/init.js
+++ b/init.js
@@ -1,16 +1,17 @@
-var mongoose = require('mongoose')
+const mongoose = require('mongoose')
+const ejson = require('mongodb-extended-json')
 
 // Configuration
 var config = require('./config')
 
-// Database
+// Load approved seed files from seeds/ into the Database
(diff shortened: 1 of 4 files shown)


jkeat:
@ti2ger92 Nice! I need to take a closer look before merging but I think it's a good approach to use for https://github.com/UPchieve/server/issues/266

Re: your Slack message about importing all JSON in the seeds folder,
I think for now we should keep this second layer control over imports so that the process is explicit and flexible.

ti2ger92:
OK, are you sure you wouldn't prefer the method from PR 271? It also keeps control in the file its self, but abstracts all code around a single metadata json structure

PR #280, opened 2020-04-04
seed users in javascript and remove mongoimport

ti2ger92:
Links
-----

Description
-----------
- For general improvement and to allow the use of a Mongo container, we remove dependency on mongo command line tools, and instead do user imports with Mongoose commands. This also maintains the same installation structure for users.


package.json
```diff
@@ -73,7 +73,8 @@
     "socket.io": "^2.2.0",
     "twilio": "^3.28.1",
     "ua-parser-js": "^0.7.21",
-    "validator": "^6.2.0"
+    "validator": "^6.2.0",
+    "mongodb-extended-json": "^1.11.0"
```
austinhallock, line 77: shouldn't this just be a dev dependency?

dbutils/seed-test-users.js
```diff
@@ -0,0 +1,36 @@
...
+var db = mongoose.connection
+db.on('error', console.error.bind(console, 'connection error:'))
+db.once('open', function() {
+  console.log('Connected to database')
+  let promises = []
+  const usersCollection = db.collection('users')
+  const seedData = require('../seeds/test_users.json')
+  // use Extended JSON to handle formats like "$date" in json files
+  const deserializedSeedData = ejson.deserialize(seedData)
+  deserializedSeedData.forEach(record => {
```
austinhallock, line 18: it'd probably be overkill for just this, but if this sort of pattern shows up a lot, using something like [bluebird](https://github.com/petkaantonov/bluebird) or [aigle](https://github.com/suguru03/aigle) simplifies lines 18-28 to

```js
Promise.map(deserializedSeedData, record => {
  return usersCollection.replaceOne({ email: record.email }, record, {upsert: true});
})
```

it also has the added benefit of being able to set concurrency, which is nice if you're dealing with a really large dataset where you don't want to have thousands of queries running at once (you can batch it X at a time)

package.json
ti2ger92: Right, sounds good.

dbutils/seed-test-users.js
ti2ger92: This actually replaces lines 20-24, doesn't it? I tried to tighten up my code, but lint wouldn't let me get away with it without extra line breaks anyway, so I think this procedural pattern is pretty readable for this case.

austinhallock: It replaces the having to keep an array of promises (so that whole loop) & the promise.all. Just pointing out bluebird / aigle as a nifty tool to make the code a little cleaner.

though i'm just a random volunteer, so my opinion doesn't matter a whole lot :P

ti2ger92: I'm a random volunteer too

PR #284, opened 2020-04-06
Faster email verification

jkeat:
Description
-----------
- Don't wait for welcome email to send before responding to registration request
- Don't wait for verification email to send before responding to verification re-send request
- Use async/await
- Remove `crypto` library (npm package is deprecated and a backward-compatible version is now built into node)


controllers/VerificationCtrl.js
```diff
@@ -1,108 +1,79 @@
...
-          )
+  initiateVerification: async function({ userId }, callback) {
+    // Find the user to be verified
+    const user = await User.findOne({ _id: userId })
+      .select('verified email')
+      .lean()
+      .exec()
+
+    if (!user) {
+      throw new Error('No account with that id found')
```
treykris, line 16: @jkeat,

Do we want to wrap all of this in a `try..catch` block and handle the errors to Sentry here or in the `auth/register` route? I don't think these errors are caught anywhere

controllers/VerificationCtrl.js
```diff
@@ -1,108 +1,79 @@
...
+    }
+
+    await User.updateOne({ _id: user._id }, userUpdates)
+
+    MailService.sendWelcomeEmail({
+      email: user.email,
+      firstName: user.firstname
+    })
+
+    return user
```
treykris, line 77: @jkeat,

I think we should be good not returning the `user` since it doesn't seem like we're using the return value in the route that calls this function (`/verify/confirm`)

controllers/VerificationCtrl.js
jkeat: Since these are expected errors we should try not to capture them with Sentry and instead handle them where needed

Agreed that the error handling in the `auth/register` route isn't great. I think that whole route should be re-written in a future PR

PR #285, opened 2020-04-07
Faster login

jkeat:
Description
-----------
- Refactor our passport LocalStrategy
    - get user from db with lean method
    - make verifyPassword a static User method
    - use async/await
- Remove unused `express-flash` middleware (wasn't slowing anything down, just totally unused)

### DIFF
--- a/models/User.js
+++ b/models/User.js
@@ -539,8 +539,8 @@ userSchema.virtual('isOnboarded').get(function() {
   return this.availabilityLastModifiedAt && isCertified
 })
 
-userSchema.statics.verifyPassword = async (candidatePassword, userPassword) => {
-  const isCorrectPassword = await new Promise((resolve, reject) => {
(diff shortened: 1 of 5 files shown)


models/User.js
```diff
@@ -553,6 +539,20 @@ userSchema.virtual('isOnboarded').get(function() {
...
+    bcrypt.compare(candidatePassword, userPassword, (error, isMatch) => {
+      if (error) {
+        return reject(error)
+      }
+
+      return resolve(isMatch)
+    })
+  })
+
+  return isCorrectPassword
```
dmnisson, line 553: Do we need this extra local variable `isCorrectPassword` if we're just going to return the `Promise` right away?

jkeat: I was trying to make it more readable, but I agree it's unnecessary and simpler to just return a promise

PR #291, opened 2020-04-10
Cordova link

treykris:
Links
-----
- Web repo PR: https://github.com/UPchieve/web/pull/377

Description
-----------
- Support for mobile app and push notifications


router/mobile/index.js
```diff
@@ -0,0 +1,66 @@
...
+      JSON.parse(process.env.FIREBASE_PRIVATE_KEY_JSON)
+    )
+  })
+
+  app.use('/setcookie', function(req, res, next) {
+    res.cookie('mobile_cookie', '1', { maxAge: 3600 * 24 * 365 * 10 })
+    res.redirect(302, 'http://localhost:12380?redirected')
+  })
+
+  app.use('/pushtest', function(req, res, next) {
```
austinhallock, line 16: rm this route, that was just for testing :)

models/User.js
```diff
@@ -193,6 +193,8 @@ var userSchema = new mongoose.Schema(
     },
     zipCode: String,
     studentPartnerOrg: String,
+
+    hasSentPushTokenRegister: { type: Boolean, default: false },
```
austinhallock, line 197: the drawback of doing this here is if a user gets a new phone (or has 2 phones) it'll be true when it shouldn't be. i'd store in localStorage on the device

router/mobile/index.js
```diff
@@ -0,0 +1,15 @@
...
+
+module.exports = function(app) {
+  admin.initializeApp({
+    projectId: 877923781231, // TODO: move to config
+    credential: admin.credential.cert(
+      JSON.parse(process.env.FIREBASE_PRIVATE_KEY_JSON)
+    )
+  })
+
+  app.use('/setcookie', function(req, res, next) {
```
austinhallock, line 11: can you add a comment `// used in native app to workaround iOS 3rd party cookie limitation`?

services/PushTokenService.js
```diff
@@ -0,0 +1,53 @@
...
+    }
+  })
+}
+
+module.exports = {
+  sendVolunteerJoined: async function(type, subTopic, tokens) {
+    const data = {
+      title: 'A volunteer joined!',
+      text: 'A volunteer has joined your session!',
+      dataStr: JSON.stringify({
```
austinhallock, line 45: i might have been the one who did the dataStr part, but it's probably more intuitive to pass in a data obj here (unstringified), then on line 10, use the obj and on line 30 use `JSON.stringify(data)`

services/PushTokenService.js
```diff
@@ -0,0 +1,53 @@
...
+  })
+}
+
+module.exports = {
+  sendVolunteerJoined: async function(type, subTopic, tokens) {
+    const data = {
+      title: 'We found a volunteer!',
+      text: 'Start chatting with your coach now.',
+      data: {
+        path: `/session/${type}/${subTopic}`
```
jkeat, line 46: Add the session id?

router/mobile/index.js
```diff
@@ -0,0 +1,16 @@
+const admin = require('firebase-admin')
+
+module.exports = function(app) {
+  admin.initializeApp({
+    projectId: 877923781231, // TODO: move to config
+    credential: admin.credential.cert(
+      JSON.parse(process.env.FIREBASE_PRIVATE_KEY_JSON)
```
jkeat, line 7: @austinhallock is this set up on production?

austinhallock: probably not, rob set it up on mobile-staging

jkeat: i think we're gonna wrap this in a conditional that checks env to allow it to run locally

jkeat: but that'll probably cause errors elsewhere when we try to use uninitialized firebase

PR #296, opened 2020-04-16
New user action - account creation

treykris:
Links
-----
- Issue: https://upchieve.monday.com/boards/438837796/pulses/531149708

Description
-----------
- Added a user action for when new accounts are created
- Changed `actionType` "PROFILE" to "ACCOUNT"
- Fixed issue where the user id was not being saved when updating their profile
- Fixed issue where `UserActionCtrl.updatedProfile` was being triggered when a new account was created - it's now set to only update from incoming requests from `/profile`


dbutils/migrate-action-type-profile.js
```diff
@@ -0,0 +1,23 @@
+const mongoose = require('mongoose')
```
jkeat, line 1: Thanks for adding this script!

PR #303, opened 2020-04-18
Fix availability

dmnisson:
Links
-----
- Issue: https://github.com/UPchieve/web/issues/360

Description
-----------
- Availability objects with no times selected (all set to `false`) are added to the test user seeds
- Call to `user.availability.toObject` is removed from `CalendarCtrl.clearSchedule`—`req.user.availability` is a plain JavaScript object since #278 was merged in, so this caused the `before` hook to fail in the e2e tests


jkeat:
Copying this from a Slack thread:
> in init.js, we currently aren't using our mongoose models, we're just adding the seed data directly to the db. If we move to using the mongoose User model (user = new User(seedData); user.save();) it would give us the default availability for free (along with a bunch of other user props that have defaults defined)

@dmnisson adding the availabilities to the seed users is a quick fix, but I think it'll be better in the long term to make our init.js use the mongoose User model instead.

We still need the fix you made in the `clearSchedule` method, though, so I'll leave this open for that!

PR #306, opened 2020-04-18
Keep same email verification token

jkeat:
Description
-----------
- don't generate new verification code for each email
    - rationale: users sometimes send multiple verification emails and have trouble finding the right link
- don't return user's email from initiateVerification (not used)


controllers/VerificationCtrl.js
```diff
@@ -8,7 +8,7 @@ module.exports = {
   initiateVerification: async function({ userId }, callback) {
     // Find the user to be verified
     const user = await User.findOne({ _id: userId })
```
treykris, line 10: @jkeat,

Would it be worth it to pass the `user` from `req.user` to this function? We might be able to skip this query

jkeat: Updated!

Still have to query the db for the user's verificationToken in most cases (verificationToken has "select: false")

treykris: Ah I see, that makes sense!

PR #302, opened 2020-04-17
Re-opened: Setup Typescript Bull Redis Queue

jkeat:
Links
-----
- Issue: https://upchieve.monday.com/boards/438837796/pulses/518195818
- Redis: https://redis.io/topics/quickstart
- Bull: https://github.com/OptimalBits/bull

Description
-----------
- adds a typescript bull worker queue
- redis db expected to be local
- adds a new `.eslintrc.js` config to handle typescript files, moves the current rules to `eslint-javascript.js`
- updates `dbconnect` to return a promise as well as take a callback, also uses the default mongoose connection

Developer self-review checklist
-------------------------------
- [ ] Potentially confusing code has been explained with comments
- [ ] No warnings or errors have been introduced; all known error cases have been handled
- [ ] Any appropriate documentation (within the code, README.md, docs, etc) has been updated
- [ ] All edge cases have been addressed

PR #307, opened 2020-04-20
Notify Volunteers through Job Queue

ericyliu:
Links
-----
- Issue: https://upchieve.monday.com/boards/438837796/pulses/518195818
- Dependency PRs: https://github.com/UPchieve/server/pull/302

Description
-----------
- notifies volunteers (tutors) through the job queue
- [Note] need to test this some more


ericyliu:
@jkeat how do you feel about starting to use ts-node (typescript) for the main api server? I didn't make those changes yet, so I have to hardcode the job name when adding it through the api

i can convert files one at a time, so I only have to convert the `services/twilio.js` file at first

jkeat:
> @jkeat how do you feel about starting to use ts-node (typescript) for the main api server? I didn't make those changes yet, so I have to hardcode the job name when adding it through the api
> 
> i can convert files one at a time, so I only have to convert the `services/twilio.js` file at first

That sounds good!

worker/jobs/notifyTutors.ts
```diff
@@ -0,0 +1,22 @@
...
+  sessionId: string;
+}
+
+export default async (job: Job<NotifyTutorsJobData>): Promise<void> => {
+  const { sessionId } = job.data;
+  const session = await Session.findById(sessionId);
+  if (!session) return log(`session ${sessionId} not found`);
+  const filled = SessionService.isSessionFilled(session);
+  if (filled) return;
+  job.queue.add(Jobs.NotifyTutors, job.data, { delay: smsTimeout });
```
jkeat, line 19: Does this recursively notify until the session is filled?

ericyliu: yep, will notify once every smsTimeout

services/twilio.js
```diff
@@ -150,6 +155,10 @@ function sendVoiceMessage(phoneNumber, messageText) {
 
   // initiate call, giving Twilio the aforementioned URL which Twilio
   // opens when the call is answered to get the TwiML instructions
+  if (!twilioClient) {
+    console.log('Twilio client not loaded.')
+    return Promise.resolve()
+  }
```
ericyliu, line 161: moved the twilioClient checks to right before sending the twilio message so we can still have a log of the API trying to send the message

PR #313, opened 2020-04-23
IP address user actions

treykris:
Description
-----------
- Add an IP address to all user actions


middleware/add-user-action.js
```diff
@@ -4,16 +4,17 @@ const Sentry = require('@sentry/node')
 function addUserAction(req, res, next) {
   if (req.user) {
     const { _id } = req.user
+    const { ip: ipAddress } = req
```
treykris, line 7: @jkeat,

There's places where you might see me do `const ipAddress = req.ip` and I've just been following the convention of the rest of the function to destructure or not destructure.

I think we should add an eslint rule for only allowing destructuring or not for consistency

jkeat: Good idea, could you file an issue for that?

PR #321, opened 2020-04-28
Notification schedule

jkeat:
Links
-----
- Dependency PRs: https://github.com/UPchieve/server/pull/307

Description
-----------
- Send notifications according to a schedule in `config.js` (array of delay times in ms)
- Volunteer prioritization (see `volunteerPriority` array)
- Send one notification at a time (simplifies a lot of things including prioritization)
- Stop notifications if session is canceled (instead of just when joined)
- Stop passing isTestUser through 5 different functions; instead, just don't send any regular notifs for test user requests
- Add "[TEST USER]" flag to failsafe texts
- Misc code cleanup (removed some unused functions)

TODO
- cache recently notified volunteers to avoid double-texting people
- test more


worker/jobs/notifyTutors.ts
```diff
@@ -9,16 +8,28 @@ import { log } from '../logger';
...
+      { sessionId, notificationSchedule },
+      { delay }
+    );
+  try {
+    const volunteerNotified = await TwilioService.notifyVolunteer(session);
+    if (volunteerNotified) log(`Volunteer notified: ${volunteerNotified._id}`);
+    else log('No volunteer notified');
+  } catch (error) {
+    log(error);
+  }
```
ericyliu, line 34: actually instead of this try catch here, could you add it to `worker/jobs/index.ts: addJobProcessors` so we dont have to wrap each script in try/catch?

ericyliu: that way we can actually get rid of the try catch around the other script as well

jkeat: good idea 👍

PR #325, opened 2020-04-29
check for availabilityLastModifiedAt before setting it

treykris:
Description
-----------
- Ignore setting `availabilityLastModifiedAt` if the user doesn't have it

### DIFF
--- a/worker/jobs/updateElapsedAvailability.ts
+++ b/worker/jobs/updateElapsedAvailability.ts
@@ -20,7 +20,8 @@ export default async (): Promise<void> => {
         );
 
         volunteer.elapsedAvailability += newElapsedAvailability;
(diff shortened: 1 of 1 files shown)


worker/jobs/updateElapsedAvailability.ts
```diff
@@ -20,7 +20,8 @@ export default async (): Promise<void> => {
         );
 
         volunteer.elapsedAvailability += newElapsedAvailability;
```
treykris, line 22: @jkeat ,

A nice little nugget that we had added here is that calculateElapsedAvailbility will return 0 if the user isn't fully onboarded

jkeat: Perfect!

PR #326, opened 2020-04-29
Ban users on non-US signup

jkeat:
Links
-----
- Issue: https://upchieve.monday.com/boards/438837796/pulses/547926029

Description
-----------
- Add `User.isBanned` (shadow banning)
    - A banned user's session request will not send notifications or show up in the waiting list
- Add `User.banReason` (enum including future ban reasons)
- Ban users on signup if their IP is outside the US (using free tier of http://ipwhois.io/)
    - 1.5 second timeout on requests. In the case of a timeout (or if we go over the API limit) we default to not banning the user


austinhallock:
if api limit ever becomes an issue, this is what i've used: https://github.com/bluesmoon/node-geoip

dataset probably isn't as accurate down to the city level, but good enough for country

models/Session.js
```diff
@@ -191,7 +191,11 @@ sessionSchema.statics.getUnfulfilledSessions = function(cb) {
   }
 
   return this.find(queryAttrs)
-    .populate({ path: 'student', select: 'firstname isVolunteer isTestUser' })
+    .populate({
+      path: 'student',
+      match: { isBanned: false },
```
jkeat, line 196: this doesn't work the way i thought it did, fixing this

jkeat: fixed

PR #327, opened 2020-04-29
Use Mongoose for good seed data

ti2ger92:
Description
-----------
- After seeing some issues compounded by missing defaults, it made sense to use Mongoose for seeding to ensure defaults are populated. Unfortunately, cypress shell still produces similar results, but this is probably a step in the right direction.


ti2ger92:
Just to clarify my comment above and from standup, this change does not introduce any bugs, and does seem to create better seed data (with defaulted structures rather than missing structures), it just doesn't fix the bug I was hoping it would. As such, I do suggest merging it.

PR #328, opened 2020-04-29
Add biology

treykris:
Links
-----
- Issue: https://upchieve.monday.com/boards/438837796/pulses/544408746
- Web PR: https://github.com/UPchieve/web/pull/393

Description
-----------
- Add biology as a certification

- [] No warnings or errors have been introduced; all known error cases have been handled
- [] Any appropriate documentation (within the code, README.md, docs, etc) has been updated


controllers/TrainingCtrl.js
```diff
@@ -13,7 +13,8 @@ var numQuestions = {
   calculus: 3,
   planning: 4,
   essays: 3,
-  applications: 2
+  applications: 2,
+  biology: 2
```
treykris, line 17: @jkeat,

I'm unsure about the number here - I just put 2 here arbitrarily

jkeat: i checked the /edu panel on prod and there are 11 biology subcategories  -- let's go with just 1, i think 22 Qs would be too long

jkeat:
Later, when we add science/biology as a subject card, we can also re-add & update the "coming soon" dashboard card

PR #329, opened 2020-04-29
IpAddress collection and service

jkeat:
Links
-----
- Issue: https://upchieve.monday.com/boards/438837796/pulses/547926088
- Dependency PR: https://github.com/UPchieve/server/pull/326

Description
-----------
- Add IpAddress collection
- Track user IP on session creation
- If banned user uses non-banned IP, ban that IP
- If non-banned user uses banned IP, ban that user (..tongue twister)
- remove `ipBlacklist` & related middleware


services/IpAddressService.js
```diff
@@ -0,0 +1,34 @@
...
+    .exec()
+
+  if (existingIpAddress) return existingIpAddress
+
+  const newIpAddress = await new IpAddress({ ip: ipString }).save()
+  return newIpAddress
+}
+
+module.exports = {
+  record: async (user, ipString) => {
```
jkeat, line 17: This function is pretty hairy/multipurposed, would appreciate any restructuring ideas

router/api/session.js
```diff
@@ -14,7 +15,13 @@ module.exports = function(router, io) {
     const data = req.body || {}
     const sessionType = data.sessionType
     const sessionSubTopic = data.sessionSubTopic
-    const user = req.user
+    const { user, ip } = req
+
+    try {
+      await IpAddressService.record(user, ip)
```
jkeat, line 21: might make this middleware and just use it on `/session/new` for now

services/IpAddressService.js
treykris: Hmmm, I can't think of anything worthwhile outside of having another function take in the `userIpAddress` and `user` and doing the check for banned IP address or user there

services/IpAddressService.js
```diff
@@ -0,0 +1,75 @@
...
+}
+
+module.exports = {
+  getIpWhoIs,
+
+  record: async (user, ipString) => {
+    const userIpAddress = await findOrCreateIpAddress(ipString)
+
+    // Ban IP if user banned
+    if (user.isBanned && IpAddress.status === IP_ADDRESS_STATUS.OK)
```
treykris, line 51: typo: `IpAddress.status` -> `userIpAddress.status`

services/IpAddressService.js
```diff
@@ -0,0 +1,75 @@
...
+      await IpAddress.updateOne(
+        { _id: userIpAddress._id },
+        { $set: IP_ADDRESS_STATUS.BANNED }
```
treykris, line 54: Oh, the status property!
 `{ $set: {status: IP_ADDRESS_STATUS.BANNED } }`

services/IpAddressService.js
```diff
@@ -0,0 +1,75 @@
...
+      )
+
+    // Ban user if IP banned
+    if (userIpAddress.status === IP_ADDRESS_STATUS.BANNED && !user.isBanned) {
+      user.isBanned = true
```
treykris, line 59: It should be alright to remove this right?

jkeat: This is the thing I mentioned that feels hacky but does technically work -- this updates the user stored on req.user so that the ban is recognized during the session creation. i.e. when SessionCtrl.create() is called, req.user.isBanned has the correct value

services/IpAddressService.js
jkeat: Ooh good catch!

services/IpAddressService.js
jkeat: It'd probably be better to return the updated user object or something like that

services/IpAddressService.js
jkeat: 👍 👍 👍

services/IpAddressService.js
jkeat: Yeah, like "banIfNecessary".. Let's just keep it big for now and split it up later if needed

services/IpAddressService.js
treykris: Oooh, okay I get it! That is a tough one

Yeah, getting the updated user back will probably feel less hacky. I also don't see many problems with this either imo. I mean I guess as long as we don't go crazy with mutating the req.user obj in different places? Idk

PR #335, opened 2020-05-04
Change Config to Typescript

ericyliu:
Description
-----------
- adds typing for config
- changes `config.example.js` to typescript
- adds runtime typechecking for config through [runtypes](https://github.com/pelotom/runtypes)


config.ts
```diff
@@ -0,0 +1,116 @@
+// Server configuration
```
jkeat, line 1: should this be tracked?

ericyliu: .. no good catch, my bad

tsconfig.json
```diff
@@ -2,5 +2,6 @@
   "compilerOptions": {
       "esModuleInterop": true,
       "allowUmdGlobalAccess": true,
-  }
+      "outDir": "build",
+  },
```
jkeat, line 6: @ericyliu quick typescript question -- when do builds come into play? It looks like a regular run (`npm run start`) doesn't involve building to this dir

ericyliu: Yeah that's a good question. I realized when running `tsc --watch` it'll compile the files and generate js files at the same location as the ts ones. This is so all the generated js files go into a specific folder. It doesn't come into play when running it in prod

ericyliu: Yeah that's a good question. I realized when running `tsc --watch` it'll compile the files and generate js files at the same location as the ts ones. This is so all the generated js files go into a specific folder. It doesn't come into play when running it in prod

PR #343, opened 2020-05-08
Derived user models (server)

treykris:
Links
-----
Web repo PR: https://github.com/UPchieve/web/pull/441

Description
-----------
- Created a base user model that Student and Volunteer now inherit from
- Added a discriminator key to the User model
- DB util to add `type` property to all users

TODO:
- ~~Update tests~~ See potential draft: https://github.com/UPchieve/server/pull/349
- ~~Change our User update queries to `Student` or `Volunteer` model~~ See https://github.com/UPchieve/server/pull/346


models/User.ts
```diff
@@ -0,0 +1,192 @@
...
+const baseUserSchemaOptions = {
+  /**
+   * https://mongoosejs.com/docs/discriminators.html#discriminator-keys
+   * The discriminator key is used to discern the different inherited models. The value of the disciminatorKey
+   * is the property that is added to a model and resolves to that type of model e.g:
+   * new Student()   --> type: "Student"
+   * new Volunteer() --> type: "Volunteer"
+   *
+   **/
+  discriminatorKey: 'type',
```
treykris, line 16: The value for the `discriminatorKey` can be any string. It's basically an identifier for the type of model that's doing the inheriting. Mongoose by default sets the `discriminatorKey` value to `__t`, but here we're setting it to `type`. This creates a `type` property on every model inheriting from baseUserSchema that resolves to the type of model it is

Example screenshots:
1.
Screen Shot 2020-05-08 at 1 19 09 PM

2.
Screen Shot 2020-05-08 at 1 19 04 PM

models/Student.ts
```diff
@@ -0,0 +1,51 @@
...
+  };
+};
+
+// Placeholder method to support asynchronous profile parsing
+studentSchema.methods.getProfile = function(cb) {
+  cb(null, this.parseProfile());
+};
+
+// Use the user schema as the base schema for Student
+const Student = User.discriminator('Student', studentSchema);
```
treykris, line 49: The discriminator key (defined in User.ts) value is set here ex: `type: "Student"`
Should we have this lowercased? to `type: student`

Wasn't sure if there was a certain type of convetion because they are capitalized here: https://mongoosejs.com/docs/discriminators.html#the-model-discriminator-function

but it works just fine lowercased as well

jkeat: i think capitalized is good!

jkeat:
@treykris looks like there's a merge conflict, probably related to the new enum option for banReason

treykris:
Updated the models!

jkeat:
@treykris Looks like there are some errors with the unit tests related to imports and/or type defs?

ti2ger92:
Unfortunately, on my Mac, this breaks e2e tests: signup-tests and student-partner-tests. I didn't run every test, but others may be broken (though session-tests is not currently active). Unless I'm missing something, this is a good reason to merge [365](https://github.com/UPchieve/server/pull/365)

jkeat:
> Unfortunately, on my Mac, this breaks e2e tests: signup-tests and student-partner-tests. I didn't run every test, but others may be broken (though session-tests is not currently active). Unless I'm missing something, this is a good reason to merge [365](https://github.com/UPchieve/server/pull/365)

still need to fix a few things, work in progress

router/auth/index.js
```diff
@@ -466,34 +466,18 @@ module.exports = function(app) {
...
+    try {
+      await ResetPasswordCtrl.finishReset({
+        email,
+        password,
+        token
+      })
+      return res.sendStatus(200)
+    } catch (err) {
+      res.status(500).json({
+        err: err.message
```
treykris, line 478: @jkeat 

When passing the error into `next(err)` we get that huge error response showing on the client-side similar to the one that Mark pinged us on via email. this should help a bit with debugging, but unsure if we should show certain errors to the client?

jkeat: Yeah we really shouldn't be sending the stack trace to the client ever, and definitely shouldn't be accidentally rendering the stack trace

jkeat: Thinking we should just capture the error with Sentry and then always return a vague "Error resetting password" message

controllers/ResetPasswordCtrl.js
```diff
@@ -60,44 +60,26 @@ module.exports = {
...
-      ],
-      callback
-    )
+      if (!user) throw new Error('No user found with that password reset token')
+
+      if (user.email !== email)
+        throw new Error('Email did not match the password reset token')
+
+      user.passwordResetToken = undefined
+      user.password = await user.hashPassword(password)
```
jkeat, line 79: this is fine for now but in general let's move away from having non-lean mongoose models, even in situations like this

controllers/ResetPasswordCtrl.js
```diff
@@ -60,44 +60,26 @@ module.exports = {
...
-
+  finishReset: async function({ email, password, token }) {
     // make sure token is a valid 16-byte hex string
     if (!token.match(/^[a-f0-9]{32}$/)) {
       // early exit
-      return callback(new Error('Invalid password reset token'))
+      throw new Error('Invalid password reset token')
     }
 
-    async.waterfall(
```
jkeat, line 73: awesome job simplifying this!

PR #348, opened 2020-05-11
Refactor register route

treykris:
Links
-----
- Web repo PR: https://github.com/UPchieve/web/pull/416

Description
-----------
- Refactored `/register` into `register/student` and `register/volunteer`


router/auth/index.js
```diff
@@ -150,8 +159,113 @@ module.exports = function(app) {
...
+      college,
+      phone,
+      favoriteAcademicSubject,
+      terms,
+      referredByCode,
+      firstName,
+      lastName
+    } = req.body
+
+    if (!terms) {
```
treykris, line 247: Would love ideas on how to refactor line `247 ~ 266` better since this is also used in the `register/student` route as well

router/auth/index.js
```diff
@@ -176,155 +290,61 @@ module.exports = function(app) {
...
-          isVolunteer: true
-        })
-      } else if (!highSchoolProvided) {
-        // Don't look up high school for students who didn't provide one (it's not required for certain partner orgs)
-        return resolve({
-          isVolunteer: false
-        })
-      }
+      // req.login loses its `this` binding when passed into promisify causing `this` not to point to `req`
+      const loginUser = promisify(req.login.bind(req))
```
treykris, line 315: This is also in `register/student` as well, might be better as a function?

router/auth/index.js
```diff
@@ -176,155 +286,58 @@ module.exports = function(app) {
...
-    }
+    const referredBy = await UserCtrl.checkReferral(referredByCode)
+
+    const volunteer = new Volunteer({
+      email,
+      isVolunteer: true,
+      registrationCode: code,
+      volunteerPartnerOrg,
+      college,
+      phonePretty: phone,
```
treykris, line 297: I still have `phonePretty` here defined, but on the Volunteer schema only `phone` is defined

models/User.ts
```diff
@@ -140,35 +140,16 @@ const baseUserSchema = new mongoose.Schema(
...
-  bcrypt.genSalt(config.saltRounds, function(err, salt) {
-    if (err) {
-      cb(err);
-    } else {
-      bcrypt.hash(password, salt, cb);
-    }
-  });
+baseUserSchema.methods.hashPassword = async function(
+  password
+): Promise<Error | string> {
```
treykris, line 145: @ericyliu 

Is this the correct way to do this?

tests/unit/models/User.test.js
```diff
@@ -435,10 +222,10 @@ test('Phone format matches', t => {
   t.is(goodUser.phonePretty, '555-555-5555')
 })
 
-test('Setting phone to null', t => {
-  goodUser.phone = null
-  t.is(goodUser.phonePretty, null)
-})
+// test('Setting phone to null', t => {
```
treykris, line 225: @jkeat 

Is there an instance where we want the phone number to be null? Like a volunteer absolutely not wanting to get texts

models/User.ts
```diff
@@ -140,35 +140,16 @@ const baseUserSchema = new mongoose.Schema(
...
+  try {
+    const salt = await bcrypt.genSalt(config.saltRounds);
+    const hash = await bcrypt.hash(password, salt);
+    return hash;
+  } catch (error) {
+    throw new error(error);
+  }
 };
 
 baseUserSchema.statics.verifyPassword = (
```
treykris, line 155: @ericyliu 

Aw man, GitHub is not showing the differences here! but I curious if this was correct as well?
https://github.com/UPchieve/server/blob/0a5bec5de16a680b943b07384e923c81021c3081/models/User.ts#L158

I'm not sure if it's always `Promise<boolean>` in this case

models/User.ts
jkeat: It seems like there's some debate about the best way to handle errors in typescript, especially expected ones (this one isn't exactly expected, though).

In this case, this approach seems fine to me 👍

interesting read: https://medium.com/inato/expressive-error-handling-in-typescript-and-benefits-for-domain-driven-design-70726e061c86

models/Volunteer.ts
```diff
@@ -120,7 +120,9 @@ const volunteerSchema = new mongoose.Schema(
     },
     phone: {
       type: String,
-      required: true
+      required: true,
+      set: (v): string => v.replace(/-/g, ''),
```
jkeat, line 124: Are there any places that are using this? i.e. doing something like:
`someUser.phone = '1-608-629-7038'; someUser.save()`
Where the phone number provided by the client has hyphens?

If not, I think we should ditch the setter

router/auth/index.js
```diff
@@ -104,23 +108,20 @@ module.exports = function(app) {
...
+      password,
+      studentPartnerOrg,
+      partnerUserId,
+      highSchoolId: highSchoolUpchieveId,
+      zipCode,
+      terms,
+      referredByCode,
+      firstName,
+      lastName
+    } = req.body
```
jkeat, line 124: What about putting ~everything past this point in a controller method so you can just do:

```
try {
  const student = await UserCtrl.createStudent({<all the user props + IP>})
  return res.json({
    user: student
  })
} catch (error)
  return res.status(422).json({ error})
}
```

Same goes for the volunteer endpoint

models/Volunteer.ts
treykris: Ditched the setter! 🙃

router/auth/index.js
treykris: Updated!

PR #349, opened 2020-05-11
Jest tests

treykris:
Links
-----
- Dependency PRs: _links to PRs that must be merged prior this one (or remove this line)_

Description
-----------
- Migrated AVA to Jest
- Added [supertest](https://www.npmjs.com/package/supertest) to do integration tests on our routes
- Jest runs MongoDB in memory using [jest-mongodb](https://github.com/shelfio/jest-mongodb)
- Integration tests on the student register route
- Express related functionality was placed into `app.ts`

TODO:
- Integration tests on the volunteer register route
- Make assertions on the mocked mail service to ensure that it has been called
- Fix WebSocket connection error that shows when trying to connect to a port already in use (3001). [this only happens when the server is already running]
- See https://github.com/UPchieve/server/blob/d7fae7e76d67d25c1fd0d23fdd7b354b08eb8ca0/app.ts#L66 before `expressWs(app,server)` was passed in
- Check if all babel dependencies are needed

Note:
In order to use `supertests` it needs to take an instance of the express app. This also allows us to run our tests against our routes without our server running. I tried to avoid having to separate out express from `main.ts`, but since `main.ts` was starting the server and setting a db connection to `upchieve`, when the tests would run they would mutate the local database instead of the in-memory database


package.json
```diff
@@ -81,20 +82,29 @@
...
+    "babel-jest": "^26.0.1",
     "eslint": "^6.8.0",
     "eslint-plugin-import": "^2.20.2",
+    "jest": "^26.0.1",
     "mongodb-extended-json": "^1.11.0",
     "nodemon": "^1.18.10",
     "nyc": "^13.3.0",
     "prettier-standard": "^9.1.1",
+    "supertest": "^4.0.2",
+    "ts-jest": "^25.5.1",
```
jkeat, line 107: You mentioned that some of these dev dependencies might be unnecessary, right?

treykris:
Updated the merge conflicts!

I will see which dependencies we can start stripping away

tests/integration/routes/register.test.ts
```diff
@@ -0,0 +1,269 @@
...
+import testHighSchools from '../../../seeds/schools/test_high_schools.json';
+import UserAction from '../../../models/UserAction';
+import { USER_ACTION } from '../../../constants';
+jest.mock('../../../services/MailService');
+
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+const registerStudent = app => request(app).post('/auth/register/student');
+
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+const createStudent = (options = {}) => {
```
ericyliu, line 15: instead of disabling the return type requirement, why not put a return type for these?

tests/integration/routes/register.test.ts
```diff
@@ -0,0 +1,269 @@
...
+beforeAll(async () => {
+  await mongoose.connect(process.env.MONGO_URL, {
+    useNewUrlParser: true
+  });
+  School.insertMany(testHighSchools);
+});
+
+afterAll(async () => {
+  await mongoose.connection.close();
+});
```
ericyliu, line 43: if we want to go with this paradigm, the mongoose connection has to be started and closed with each test file, which might take up a lot of time in the long run. There we might be a scaffolding file where you can open the mongo connection, start the jest tests, then close it after the tests are run that could alleviate this issue

tests/integration/routes/register.test.ts
```diff
@@ -0,0 +1,269 @@
...
+
+test('Student did not provide an email', async () => {
+  const studentOptions = { email: '' };
+  const newStudent = createStudent(studentOptions);
+
+  const response = await registerStudent(app)
+    .send(newStudent)
+    .set('Accept', 'application/json')
+    .expect(422);
```
ericyliu, line 71: if you're going to make this registerStudent call with `newStudent` and the set call every time, maybe abstract that out into the registerStudent function itself so you dont have code duplication in every test

`const registerStudent = (app: Express, newStudent: Student, options: { region }): Jest.Mock => { ... }`

also, do you need to pass in the app during the call? its something thats being imported at the top level of the file anyways

tests/unit/controllers/ModerationCtrl.test.js
```diff
@@ -1,50 +1,49 @@
...
 const ModerationCtrl = require('../../../controllers/ModerationCtrl')
 
-test.cb('Check incorrect email succeeds', t => {
+test('Check incorrect email succeeds', done => {
   const email = { content: 'j.@serve1.proseware.com' }
   ModerationCtrl.moderateMessage(email, function (err, res) {
-    t.truthy(res)
-    t.end()
+    expect(res).toBeTruthy()
+    done()
```
ericyliu, line 7: might be more consistent to change this to async await instead of using `done`

tests/integration/routes/register.test.ts
treykris: Sleepy me thought that I would ask Eric about this later haha, I'm not sure how to do the return type on those

tests/integration/routes/register.test.ts
treykris: Interesting! I'll look into this

tests/integration/routes/register.test.ts
treykris: Ah nice! I started to think it was too much duplication, but wasn't sure on a good way to approach it, thanks! For the return type of `options` with Jest.Mock should it just be all the object properties being returned from the function?

I remember having issues when I didn't pass in the app during each call, but it could have been something I was doing wrong - I'll revisit

seeds/schools/test_high_schools.json
```diff
@@ -6,15 +6,9 @@
         "cityNameStored": "New York",
         "stateStored": "NY",
         "isApproved": false,
-        "createdAt": {
```
ti2ger92, line 9: Why these changes?

treykris: I was getting some validation casting errors when the date was included so I just let the model use the School model resolve the dates to `date.now()`

Screen Shot 2020-05-27 at 9 33 48 PM

treykris: I just realized alternatively I do this `"createdAt": "2020-03-29T19:43:57.068Z"` and it should work fine

I think the `$date` doesn't cast the string properly when running the tests

tests/integration/routes/register.test.ts
ericyliu: `options` could just be whatever you need to pass in that will enable custom behavior when setting up the mock. I noticed you set the region for one of these instances (but not for most of them), so that could be an option. 

As for `app` it might be a scoping issue and its safer to pass it in for each instance, so it mightve been a legitimate problem you ran into

tests/integration/routes/register.test.ts
ericyliu: the syntax would be:

```
interface Student {
   ... everything that a student has
}
const createStudent = (options: Partial<Student> = {}): Student  => { ... }
```

the `Partial<Student>` means options can have any subset of a `Student`'s keys, and the return type is a `Student` because you're combining the two objects

you might want to declare the interface inside the Student model file, but that would involve changing it to a typescript file, which might be more work than you want to do right now (should definitely be where it lives in the future once that becomes a typescript file)

tests/integration/routes/register.test.ts
ericyliu: i wouldn't spend too much time on this, it's something I looked into recently and had some trouble doing for another project

PR #351, opened 2020-05-12
Fault tolerance

dmnisson:
Links
-----
- Issue: https://github.com/UPchieve/web/issues/400
- Web repo PR: https://github.com/UPchieve/web/pull/413

Description
-----------
- Instead of using rooms to manage two-user sessions, the correct socket for the desired recipient user for an event is determined by looking up the user in a map of user ObjectIDs to sockets in memory


dmnisson:
TODO implement fault-tolerant chat messaging, with indictors that app is trying to send message

services/SocketService.js
```diff
@@ -38,27 +41,33 @@ module.exports = function(io) {
...
+        userSocket => socket.id === userSocket.id
+      )
+
+      userSockets[userId].splice(socketIndex, 1)
+    },
+
+    emitToUser: function(userId, event, ...args) {
+      const sockets = userSockets[userId]
+      if (sockets && sockets.length) {
+        sockets[sockets.length - 1].emit(event, ...args)
```
jkeat, line 70: @dmnisson Would there be any downsides to emitting to all the user's current sockets? Thinking about the case where a volunteer joins on their phone and then computer.

```
for (const socket of sockets) {
  socket.emit(event, ...args)
}
```

services/SocketService.js
```diff
@@ -106,15 +113,23 @@ module.exports = function(io) {
...
         contents: message.contents,
         name: message.user.firstname,
         userId: message.user._id,
         isVolunteer: message.user.isVolunteer,
         picture: message.user.picture,
         createdAt: message.createdAt
-      })
+      }
+
+      await this.emitToOtherUser(
```
jkeat, line 126: does this need to be `await`ed?

dmnisson: No, it doesn't.

PR #354, opened 2020-05-13
added banned service providers

treykris:
Description
-----------
- Ban users that are using a service provider that we have banned


router/auth/index.js
```diff
@@ -260,9 +260,13 @@ module.exports = function(app) {
...
         if (!user.isVolunteer) {
           const {
-            country_code: countryCode
+            country_code: countryCode,
+            org
           } = await IpAddressService.getIpWhoIs(req.ip)
-          if (countryCode && countryCode !== 'US') {
+          if (
+            (countryCode && countryCode !== 'US') ||
+            config.bannedServiceProviders.includes(org)
```
jkeat, line 268: Could you make it so a user w/ banned service provider has a different ban reason?

PR #356, opened 2020-05-14
update Docker so init can run in ts

ti2ger92:
Description
-----------
- Found that docker-compose up did not run. Fixed by: removing config.js, and changing `node` to `npx ts-node` in my calls. Made same change in Readme

### DIFF
--- a/bin/docker_run.sh
+++ b/bin/docker_run.sh
@@ -1,9 +1,14 @@
 #!/usr/bin/env bash
+
+#Alter a line in default config.ts: 
+#     database: 'mongodb://localhost:27017/upchieve',
+#New line should read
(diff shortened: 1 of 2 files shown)


bin/docker_run.sh
```diff
@@ -1,9 +1,14 @@
 #!/usr/bin/env bash
+
+#Alter a line in default config.ts: 
+#     database: 'mongodb://localhost:27017/upchieve',
+#New line should read
+#     database: 'mongodb://db:27017/upchieve',
```
jkeat, line 6: Just to clarify, "db" is the name of mongo container defined in `docker-compose.yml` right?

ti2ger92: Yes indeed.

PR #363, opened 2020-05-21
Integrated math

treykris:
Links
-----
- Web repo PR: https://github.com/UPchieve/web/pull/430

Description
-----------
- Added Integrated Math 1 - 4 as a certification
- Certify a volunteer in Integrated Math 1 - 4 if the requirements are each certification is met when a volunteer passes a quiz

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -75,7 +75,7 @@ const IP_ADDRESS_STATUS = {
 
 const INTEGRATED_MATH_MAPPING = {
   integratedmathone: 'integratedMathOne',
-  integratemathtwo: 'integratedMathTwo',
+  integratedmathtwo: 'integratedMathTwo',
   integratedmaththree: 'integratedMathThree',
   integratedmathfour: 'integratedMathFour'
 }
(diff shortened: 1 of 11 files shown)


dbutils/add-integrated-math.ts
```diff
@@ -0,0 +1,121 @@
+import mongoose from 'mongoose';
+import User from '../models/User';
+import dbconnect from './dbconnect';
+
+async function upgrade(): Promise<void> {
+  try {
+    await dbconnect();
+    const result = await User.bulkWrite([
```
jkeat, line 8: perf migration script 🤩

constants.js
```diff
@@ -73,6 +73,20 @@ const IP_ADDRESS_STATUS = {
   BANNED: 'BANNED'
 }
 
+const INTEGRATED_MATH_MAPPING = {
```
jkeat, line 76: It seems like the only reason we need these mappings between lowercase and camel case is that we grab the subtopic directly from the URL on the client, and that's lowercased here:
https://github.com/UPchieve/web/blob/master/src/utils/session.js#L10

The URL should be case insensitive, so the fix isn't to remove this lowercasing. Instead we shouldn't be grabbing the subtopic from the URL at all, it should be handled internally on the client. That can be a later PR though

PR #367, opened 2020-05-27
Elapsed availability

treykris:
Description
-----------
- Moved calculateElapsedAvailability from User to UserCtrl and the respective tests
- When a volunteer updates their profile calculate their elapsed avail


controllers/UserCtrl.js
```diff
@@ -139,5 +166,51 @@ module.exports = {
 
   deleteUserByEmail: function(userEmail) {
     return User.deleteOne({ email: userEmail }).exec()
+  },
+
+  // Calculates the amount of hours between a volunteer's availabilityLastModifiedAt
+  // and the current time that a user updates to a new availability.
+  // Expects a "lean" (non-Mongoose doc) volunteer to be passed in,
+  // otherwise the volunteer needs to be coerced using the mongoose method "toObject()"
+  calculateElapsedAvailability: function(volunteer, newModifiedDate) {
```
treykris, line 175: I decided to pass in the `newModifiedDate` with the `volunteer` so that we can make this function testable

worker/jobs/updateElapsedAvailability.ts
```diff
@@ -15,7 +16,8 @@ export default async (): Promise<void> => {
     await Promise.all(
       map(volunteers, async volunteer => {
         const currentTime = new Date();
-        const newElapsedAvailability = volunteer.calculateElapsedAvailability(
+        const newElapsedAvailability = UserCtrl.calculateElapsedAvailability(
+          volunteer.toObject(),
```
treykris, line 20: `calculateElapsedAvailability` expects a regular object for the volunteer, so we can just convert mongoose docs with `toObject()`

jkeat: Now we could have this job fetch the volunteers lean, right?

treykris: Yup! We'd just have to change this to update the volunteer instead of doing `await volunteer.save();`
I can make those changes if that's a direction we'd want to go

jkeat: That's definitely the direction we want to go but let's save that for later

jkeat:
Looks good! Just need to do a bit of ~smoke testing locally with the queue

PR #371, opened 2020-05-27
Session partner status indicator

dmnisson:
Links
-----
- Web repo PR: https://github.com/UPchieve/web/pull/440

Description
-----------
- Socket events to indicate to a session participant when the partner connects or disconnects

### DIFF
--- a/services/SocketService.js
+++ b/services/SocketService.js
@@ -68,26 +68,30 @@ module.exports = function(io) {
       userSockets[userId].splice(socketIndex, 1)
 
       // update session partner on connection status
-      const currentSession = await Session.current(userId);
+      const currentSession = await Session.current(userId)
       if (currentSession) {
         this.updateConnectionStatus(currentSession, userId, false)
       }
     },
 
-    updateConnectionStatus: function(currentSession, userId, isConnectionAlive) {
+    updateConnectionStatus: function(
+      currentSession,
+      userId,
+      isConnectionAlive
+    ) {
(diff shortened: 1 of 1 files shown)


controllers/SessionCtrl.js
```diff
@@ -142,6 +142,11 @@ module.exports = function(socketService) {
         }
 
         socketService.emitSessionChange(session._id)
+        socketService.updateConnectionStatus(
+          session,
+          user._id,
+          socketService.isConnected(user._id)
+        )
```
ericyliu, line 149: do you need this if the update also happens in `connectUser`?

services/SocketService.js
```diff
@@ -62,6 +70,37 @@ module.exports = function(io) {
...
+      userId,
+      isConnectionAlive
+    ) {
+      const sessionPartner = currentSession.student._id.equals(userId)
+        ? currentSession.volunteer
+        : currentSession.volunteer._id.equals(userId)
+        ? currentSession.student
+        : null
+
+      if (sessionPartner) {
```
ericyliu, line 92: a trick you can use is do an early return if you dont plan on doing anything else in the function, i.e.
`if (!sessionPartner) return`

this means you dont have to indent the next block of code and keeps it a bit more readable imo

controllers/SessionCtrl.js
dmnisson: If the user joins a session while their socket is already connected, the `connectUser` function will not be called.

PR #372, opened 2020-05-27
Update Models To Typescript

ericyliu:
Links
-----
- Issue: _link to GitHub issue (or remove this line)_
- Web repo PR: _link to companion PR in server repo (or remove this line)_
- Dependency PRs: _links to PRs that must be merged prior this one (or remove this line)_

Description
-----------
- updates some models to typescript (didn't do all because this pr was already becoming quite big)


models/feedback.ts
```diff
@@ -1,6 +1,17 @@
-var mongoose = require('mongoose')
-
-var feedbackSchema = new mongoose.Schema({
+import { Document, model, Schema } from 'mongoose';
+
+export interface Feedback extends Document {
+  sessionId: string;
+  type: string;
+  subTopic: string;
+  responseData: { [key: string]: any };
```
ericyliu, line 7: this is what sendgrid expects (as far as I can tell, this is being used to send stuff to sendgrid)

models/feedback.ts
```diff
@@ -1,6 +1,17 @@
-var mongoose = require('mongoose')
-
-var feedbackSchema = new mongoose.Schema({
+import { Document, model, Schema } from 'mongoose';
+
+export interface Feedback extends Document {
+  sessionId: string;
+  type: string;
```
ericyliu, line 5: this might need to be more restrictive than string

models/ipAddress.ts
```diff
@@ -0,0 +1,29 @@
+import { Document, model, Schema, Types } from 'mongoose';
+import { IP_ADDRESS_STATUS } from '../constants';
+
+export interface IpAddress extends Document {
+  createdAt: Date;
+  ip: string;
+  users: Types.ObjectId[];
+  status: IP_ADDRESS_STATUS.OK | IP_ADDRESS_STATUS.BANNED;
```
ericyliu, line 8: maybe consider moving `IP_ADDRESS_STATUS` to this model file?

ericyliu:
changed model file names from uppercase start to lowercase start, to make it consistent with the rest of the project, i can revert this if we feel strongly about this change

models/message.ts
```diff
@@ -25,25 +27,25 @@ var messageSchema = new mongoose.Schema(
...
 
 messageSchema.virtual('picture').get(function() {
   // only works if user is populated
-  return this.user.picture
-})
+  return this.user.picture;
+});
 
-module.exports = mongoose.model('Message', messageSchema)
+export default model<Message>('Message', messageSchema);
```
ericyliu, line 51: not sure if I should include virtuals in the typing (i think i will need to later, when the files using the model is converted and its expecting those keys to be there)

models/feedback.ts
jkeat: hmm this isn't related to sendgrid
https://github.com/UPchieve/server/blob/master/router/api/feedback.js#L10
https://github.com/UPchieve/web/blob/master/src/views/FeedbackView.vue#L334

ericyliu: err the only way its used in the api at least is passed into SendGrid through MailService
https://github.com/UPchieve/server/blob/master/services/MailService.js#L67

PR #373, opened 2020-05-29
Admin filter tools

treykris:
Links
-----
- Web repo PR: https://github.com/UPchieve/web/pull/444

Description
-----------
- Added new filtering options for the admin panel:  filter banned users, test users, filter by session length, by the number of messages that were sent in a session and the date range of session activity


router/api/session.js
```diff
@@ -180,13 +180,75 @@ module.exports = function(router, io) {
...
     const page = parseInt(req.query.page) || 1
     const skip = (page - 1) * PER_PAGE
+    const {
+      showBannedUsers,
+      showTestUsers,
+      minSessionLength,
+      sessionActivityFrom,
+      sessionActivityTo,
+      minMessagesSent
+    } = req.query
```
ericyliu, line 190: Possibly consider using the [runtypes](https://github.com/pelotom/runtypes) library (used for importing the config) for runtime typechecking of the query

router/api/session.js
```diff
@@ -180,13 +180,75 @@ module.exports = function(router, io) {
...
+
+    // Add a day to the sessionActivityTo to make it inclusive for the activity range: [sessionActivityFrom, sessionActivityTo]
+    const inclusiveSessionActivityTo =
+      new Date(sessionActivityTo).getTime() + 1000 * 60 * 60 * 24
```
ericyliu, line 194: should magic numbers be included at the top of the file?

router/api/session.js
```diff
@@ -180,13 +180,75 @@ module.exports = function(router, io) {
...
+          }
+        },
+        {
+          $unwind: '$student'
+        }
+      ])
         .sort({ createdAt: -1 })
         .skip(skip)
         .limit(PER_PAGE)
-        .lean()
```
ericyliu, line 189: cant be lean?

treykris: Awesomely enough using aggregation returns regular objects

router/api/session.js
treykris: All of the types will always come in as strings, would runtypes still be pretty useful?

controllers/SessionCtrl.js
```diff
@@ -203,6 +203,92 @@ module.exports = function(socketService) {
...
+      const pageNum = parseInt(page) || 1
+      const skip = (pageNum - 1) * PER_PAGE
+      const oneDayInMS = 1000 * 60 * 60 * 24
+
+      // Add a day to the sessionActivityTo to make it inclusive for the activity range: [sessionActivityFrom, sessionActivityTo]
+      const inclusiveSessionActivityTo =
+        new Date(sessionActivityTo).getTime() + oneDayInMS
+
+      try {
+        const sessions = await Session.aggregate([
```
jkeat, line 227: It looks like the aggregation pipeline doesn't follow mongoose's `select: false` rules, so this ends up returning users' passwords, the whiteboard doc, etc.

Can you see if there's a way to follow the `select: false` rules, or else just project/select the necessary properties?

treykris: Ooo nice catch!

I tried searching for a way that we could still follow the `select: false` rules, but I couldn't find one. I'm guessing that the aggregation is somehow overriding the schema defined rules / might be a design choice for aggregation to always get the model in full and then explicitly `$project` or `$unset`?

I updated it to explicitly exclude the fields with `select: false`!

controllers/SessionCtrl.js
```diff
@@ -203,6 +203,104 @@ module.exports = function(socketService) {
...
+                }
+              ],
+              as: 'student'
+            }
+          },
+          {
+            $unwind: '$student'
+          },
+          {
+            $unset: [
```
treykris, line 281: Instead of `$unset` we could also do something like:

``` {
            $unwind: '$student'
          }
        ])
          .project('-student.password -student.verificationToken -whiteboardDoc')
          .sort({ createdAt: -1 })
          .skip(skip)
          .limit(PER_PAGE)
          .exec()
```

jkeat: In this case it might be better to select the fields you want rather than the ones you don't want

If we add a new private field later, we'd have to remember to update this. If we need to give the client access to a new public field, we'll know we have to update this

treykris: Ah, I see! That's a really good point

I just realized that `$project` was the better option too because the `$unset` sends back the populated student and the student isn't being used when listing out all the filtered sessions

PR #375, opened 2020-05-29
Emit latest session when user reconnects instead of current session

dmnisson:
Links
-----
- Issue: https://github.com/UPchieve/web/issues/442

Description
-----------
- Instead of emitting the current session or an empty object if the current session does not exist, the user gets their _latest_ session in the `session-change` event when their socket connects. This way the client knows when the session it was in has ended.


ericyliu:
when would the current session not be the latest session? can you visit previous sessions?

dmnisson:
> when would the current session not be the latest session? can you visit previous sessions?

When the session ends, neither participant has a current session.

PR #383, opened 2020-06-09
Updated volunteer stats

treykris:
Description
-----------
- Use the latest message sent to calculate the total tutoring hours completed for a tutor
- For sessions that are longer than 3 hours search for the latest message was sent within 15 minutes of the message before ( a better alternative might be to ignore messages sent from the same person within that 15 minute window )

Would love any thoughts on how to improve the readability

### DIFF
--- a/controllers/UserCtrl.js
+++ b/controllers/UserCtrl.js
@@ -40,28 +40,47 @@ function isOnboarded(volunteer) {
 module.exports = {
   getVolunteerStats: async user => {
     const pastSessions = await Session.find({ volunteer: user._id })
-      .select('volunteerJoinedAt endedAt')
+      .select('volunteerJoinedAt endedAt messages')
       .lean()
       .exec()
+    const threeHoursMs = 1000 * 60 * 60 * 3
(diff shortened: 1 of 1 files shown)


controllers/UserCtrl.js
```diff
@@ -40,29 +40,46 @@ function isOnboarded(volunteer) {
...
+      // get the latest message that was sent within a 15 minute window of the message prior.
+      // Sometimes sessions are not ended by either participant and one of the participants may send
+      // a message to see if the other participant is still active before ending the session.
+      // Exclude these messages when getting the total session end time
+      if (sessionLengthMs > threeHoursMs) {
+        while (
+          latestMessageIndex > 0 &&
+          messages[latestMessageIndex].createdAt -
+            messages[latestMessageIndex - 1].createdAt >
+            fifteenMinsMs
```
jkeat, line 74: I think the lodash function findLast (https://lodash.com/docs/2.4.2#findLast) could make this cleaner

jkeat: Though maybe not since you're dealing with indexes -- this looks good to me for now

Issue #298, opened 2020-04-17
Registration endpoint is messy


jkeat:
Refactor to handle business logic outside of the router, use async/await, etc

PR #384, opened 2020-06-10
Refactor register

ti2ger92:
Links
-----
- Issue: https://github.com/UPchieve/server/issues/298

Description
-----------
- I updated the api documentation, transitioned the router to typescript, and cleaned up some of the share functions.

- I worked with the assumptions that we want to keep response codes in the router if convenient, which is why I left a lot of the stuff in the router, but I'm open to suggestions.


jkeat:
@ti2ger92 just checking, is this ready for review?

ti2ger92:
@jkeat , Sorry, but no. Last weekend the refactoring took on a  life of its own (as sometimes happens) when I tried to do all business logic outside the router and all return codes inside the router by leveraging throwables (because my first love will always be Java). Then last week my day job kicked into high gear, so I didn't get a chance to finish it. I'll try to take another pass this week, but no one should feel blocked by this, if someone has time, they should feel free to continue or ignore my work.

PR #394, opened 2020-06-18
Reference email worker

jkeat:
Links
-----
- Dependency PRs: https://github.com/UPchieve/server/pull/376

Description
-----------
- Add worker that emails unsent references
- update reference status after successful send


services/MailService.js
```diff
@@ -34,6 +34,13 @@ const sendEmail = (
   sgMail.send(msg, callback)
```
treykris, line 34: I just realized that we should probably return this so that we can do error handling for the mail functions. I think this will be an uncaught error without the return

services/MailService.js
```diff
@@ -34,6 +34,13 @@ const sendEmail = (
...
 }
 
+// @todo: use this in other MailService methods
+const buildLink = path => {
```
treykris, line 38: nice!

worker/jobs/emailReferences.ts
```diff
@@ -0,0 +1,51 @@
...
+  reference: Reference;
+  volunteer: Volunteer;
+}
+
+export default async (): Promise<void> => {
+  // @todo: don't wrap in giant try/catch?
+  try {
+    await dbconnect();
+
+    const volunteers = (await VolunteerModel.find({
```
treykris, line 20: We can also consider using aggregation here if we want. I'll post this as a reference for the future in case we decide to use it

```
    const unsent: Array<UnsentReference> = await VolunteerModel.aggregate([
      {
        $match: { 'references.status': REFERENCE_STATUS.UNSENT }
      },
      {
        $unwind: '$references'
      },
      {
        $match: { 'references.status': REFERENCE_STATUS.UNSENT }
      },
      {
        $project: {
          reference: '$references',
          volunteer: {
            firstname: '$firstname',
            lastname: '$lastname'
          }
        }
      }
    ]);
```

jkeat: Solid! gonna leave it for now but this is great for reference

PR #398, opened 2020-06-26
Test volunteer signup

treykris:
Description
-----------
- tests for the new volunteer sign up


package.json
```diff
@@ -6,8 +6,8 @@
     "start": "ts-node main",
     "dev": "ts-node-dev main",
     "dev:autolint": "nodemon main --exec 'npm run lint && node'",
-    "test": "jest",
-    "test:watch": "jest --watch",
+    "test": "jest --runInBand --detectOpenHandles --forceExit",
```
treykris, line 9: `--forceExit` flag resolves the issue of the jest tests not exiting once finished. This was causing long builds in Circle CI. Might be an issue with super test. This should be resolved in the future

https://github.com/visionmedia/supertest/issues/520

https://github.com/facebook/jest/issues/5783

.circleci/config.yml
```diff
@@ -5,7 +5,7 @@ jobs:
   build:
     docker:
       - image: circleci/node:11
-      - image: circleci/mongo:3.6-stretch-ram
+      - image: circleci/mongo:4.2.3-ram
```
jkeat, line 8: did it cause issues to remove this?

treykris: Yup, it basically starts the MongoDB process so that we can connect to the db when our tests run

PR #406, opened 2020-07-07
data for graphs

austinhallock:
I only really shared this with Aly, so not sure if you guys saw, but Rachel and I are working on [this](https://techby.org/), so our goal is to try to build stuff that helps a lot of nonprofits. You guys are the guinea pigs for this.

This code basically takes the MongoDB data and converts it to a datapoint format I'm using on our in end ScyllaDB. Basically everything is stored by a metricId and dimensionId/value with a counter and "scaled" time. So for 2020-07-07 there's a counter for a metricId+dimensionId to see how many there were for that day. Similarly there are counters for each week, month, year and all-time. ([code](https://github.com/techbyorg/impact-api/blob/master/graphql/datapoint/model.js) if you're curious)

http://localhost:3000/metrics?minTime=2020-04-01&maxTime=2020-07-01&timeScale=month grabs all of the datapoints from 4/1 to 7/1 at a time scale of months (so count per month). 

Averages, percentages, etc... are calculated on the fly by dividing one metric by another (eg total # of seconds across all sessions divided by # of sessions to get avg session length). 

Eventually the implementation of this will just be to include a TechBy library in Node and call something like `TechBy.log('session', { topic: 'math', subTopic: 'prealgebra'})` and `TechBy.log('sessionDuration', { topic: 'math', subTopic: 'prealgebra'}, 360)` (where 360 is session length in seconds). But the code in the PR would have been necessary regardless to import past data 

I also figured I'd start with this version though so you guys have something to fallback to in case we shut ever shut our end down (though we could definitely export the data as well). 

On my end, I'll query all datapoints for the past (eg https://upchieve.org/api/metrics?minTime=2018-01-01&maxTime=2020-07-07&timeScale=month) once for each timeScale (day, month, year, all), then after that I'll only need to query for the current day (https://upchieve.org/api/metrics?minTime=2020-07-07&maxTime=2020-07-07&timeScale=day) to keep the data up-to-date. I'll probably do that at an interval of once an hour. The query for just today's numbers shouldn't be too intensive on Mongo

Here's an example of the datapoints that get generated:

```
{
  "slug": "sessions",
  "name": "Sessions",
  "datapoints": [{
        "count": 274,
        "scaledTime": "MON:2020-04",
        "dimensionSlug": "all",
        "dimensionValue": "all"
      },
      {
        "count": 376,
        "scaledTime": "MON:2020-05",
        "dimensionSlug": "all",
        "dimensionValue": "all"
      },
      {
        "count": 175,
        "scaledTime": "MON:2020-06",
        "dimensionSlug": "all",
        "dimensionValue": "all"
      },
      {
        "count": 75,
        "scaledTime": "MON:2020-04",
        "dimensionSlug": "day-of-week",
        "dimensionValue": "Wed"
      },
      {
        "count": 53,
        "scaledTime": "MON:2020-04",
        "dimensionSlug": "day-of-week",
        "dimensionValue": "Thu"
      },
      {
        "count": 37,
        "scaledTime": "MON:2020-04",
        "dimensionSlug": "day-of-week",
        "dimensionValue": "Fri"
      },
      {
        "count": 22,
        "scaledTime": "MON:2020-04",
        "dimensionSlug": "day-of-week",
        "dimensionValue": "Sat"
      },
      {
        "count": 7,
        "scaledTime": "MON:2020-04",
        "dimensionSlug": "day-of-week",
        "dimensionValue": "Sun"
      },
...
```

There's the generic "all" dimension, but also counts for different dimensions like day of week, topic, etc...

Let me know if you have any questions


austinhallock:
This is a public facing route, so technically anyone can access it. IMO the data isn't sensitive since it's all just aggregate numbers, but we can put it behind a password if needed

router/metrics/helpers.js
```diff
@@ -0,0 +1,338 @@
...
+  const sessions = await Session.find({
+    createdAt: { $gte: minTime, $lte: maxTime }
+  }).select([
+    'createdAt',
+    'volunteerJoinedAt',
+    'endedAt',
+    'messages',
+    'type',
+    'subTopic'
+  ])
```
jkeat, line 121: could this be lean?

austinhallock: yeah, they can all be lean, i'll update them

PR #408, opened 2020-07-09
Fix past sessions

treykris:
Description
-----------
- Updated args to addPastSession
- Db util to add missed sessions to a user's `pastSessions`

### DIFF
--- a/services/SessionService.js
+++ b/services/SessionService.js
@@ -26,7 +26,10 @@ module.exports = {
     if (!isAdmin && !isSessionParticipant(session, endedBy))
       throw new Error('Only session participants can end a session')
 
-    await addPastSession(session.student, session._id)
+    await addPastSession({
+      userId: session.student._id,
+      sessionId: session._id
+    })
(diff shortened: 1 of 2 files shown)


services/SessionService.js
```diff
@@ -26,7 +26,10 @@ module.exports = {
     if (!isAdmin && !isSessionParticipant(session, endedBy))
       throw new Error('Only session participants can end a session')
 
-    await addPastSession(session.student, session._id)
+    await addPastSession({
+      userId: session.student._id,
```
jkeat, line 30: Hm wait so is `session.student` populated or something?

Does `session.volunteer` on line 33 need the same treatment?

treykris: I think the `._id` for an ObjectId property is exposed even if it's not populated. I'll update it to just use the `session.student` because it does look like it implies that it is populated and yup it does, nice catch!!

jkeat: Ohh I didn't realize the main change was wrapping these in an object instead of positional args. Good job finding this!

PR #405, opened 2020-07-06
Volunteer signup emails

jkeat:
Links
-----
- Dependency PRs: https://github.com/UPchieve/server/pull/376

Description
-----------
- worker for sending ready-to-coach email
- emails for
    - photo id rejected
    - approved-but-not-onboarded
    - ready to coach
- separate welcome emails for partner vs regular volunteers

todo:
- volunteers dont show up on review page unless background info completed

Developer self-review checklist
-------------------------------
- [ ] Potentially confusing code has been explained with comments
- [ ] No warnings or errors have been introduced; all known error cases have been handled
- [ ] Any appropriate documentation (within the code, README.md, docs, etc) has been updated
- [ ] All edge cases have been addressed

PR #409, opened 2020-07-10
Volunteer signup user actions

treykris:
Description
-----------
- Added user actions for `APPROVED`,`ONBOARDED`,`ADDED_REFERENCE`,`ADDED_PHOTO_ID`,`ADDED_BACKGROUND_INFO`,`DELETED_REFERENCE`
- Updated UserService tests to test that the user actions were created


router/api/volunteers.js
```diff
@@ -61,13 +61,19 @@ module.exports = function(router) {
     const { volunteerId, photoIdStatus, referencesStatus } = req.body
 
     try {
-      const { occupation, background, country } = await UserService.getUser({
+      const {
+        occupation,
+        background,
+        country,
+        references
+      } = await UserService.getUser({
```
treykris, line 69: @jkeat,

Realized that we ended up having two similar methods for getting the in our PRs `getUser` and `getVolunteer` haha

I'm thinking about refactoring and following your [approach](https://github.com/UPchieve/server/blob/0cf1fbac314cdfa4c75e3414466985e93ab06ba9/services/UserService.js#L161) in https://github.com/UPchieve/server/pull/405 because I like the idea of testing that user actions are being saved without having to worry about passing the relevant input for the `UserActionCtrl` through the input of a `UserService` method. We're less worried about getting the correct inputs upfront for the services and instead are testing that the `UserService` gets its relevant input and then gets the data that it needs to pass into `UserActionCtrl`


Current approach: Route (gets data for both the UserService and UserActionCtrl) >  pass to both `UserService`  and `UserActionCtrl`

refactored approach: Route passes relevant data for UserService > `UserService.method` gets data relevant for `UserActionCtrl`

services/UserService.js
```diff
@@ -167,7 +182,21 @@ module.exports = {
...
     }
 
-    return Volunteer.update({ _id: volunteerId }, update)
+    if (photoIdStatus === STATUS.REJECTED)
+      UserActionCtrl.rejectedPhotoId(volunteerId)
+
+    // @todo: volunteer-signup - pending merge with incoming emails pr
+    for (let i = 0; i < referencesStatus.length; i++) {
+      if (referencesStatus[i] === STATUS.REJECTED)
+        UserActionCtrl.rejectedReference(volunteerId, {
```
jkeat, line 191: for this, we'll want to add a check that makes sure it's newly rejected, right? as in, make sure this only gets triggered on initial rejection

router/reference/index.js
```diff
@@ -6,9 +6,24 @@ module.exports = function(app) {
...
-    await UserService.saveReferenceForm({ referenceId, referenceFormData })
+    const { references, _id: userId } = await UserService.getUser({
+      'references._id': referenceId
+    })
+
+    let referenceEmail
+    for (const reference of references) {
+      if (reference._id.toString() === referenceId)
+        referenceEmail = reference.email
+    }
```
jkeat, line 19: just for reference, with lodash:
```
const reference = _.find(references, (r) => r._id.toString() === referenceId)
const referenceEmail = reference.email

// or

const referenceEmail = _.chain(references).find((r) => r._id.toString() === referenceId).get('email').value()
```

PR #412, opened 2020-07-16
Wb photo upload

treykris:
Links
-----
- Web repo PR: https://github.com/UPchieve/web/pull/479

Description
-----------
- Extract images from the whiteboard doc and send to client from the admin session route


services/SessionService.js
```diff
@@ -69,5 +70,14 @@ module.exports = {
     })
       .lean()
       .exec()
+  },
+
+  getSessionPhotoUploadUrl: async sessionId => {
+    const sessionPhotoS3Key = crypto.randomBytes(32).toString('hex')
```
jkeat, line 76: small thing, but this would feel even safer if you prefixed the random key with the session ID itself, which is guaranteed to be unique (and if you do that, you could bring down the random bytes to 8 or 16)

treykris: Good decision, updated!

PR #414, opened 2020-07-16
Eslint fixes

austinhallock:
Combined the two eslint files into one. Still a little weird that the code style for ts files is pretty different from the one for js files :P but this at least makes lint work by default for all (and fixed all the errors that came up)


austinhallock:
fixing build

PR #417, opened 2020-07-18
Sg create contacts

treykris:
Description
-----------
- Add a new contact to SendGrid when a student or volunteer signs up
- Db script to add users to sendgrid contacts


services/MailService.js
```diff
@@ -137,5 +166,34 @@ module.exports = {
       { sessionId, reportedByEmail, reportMessage },
       config.sendgrid.unsubscribeGroup.account
     )
+  },
+
+  createContact: async user => {
+    const customFields = await getCustomFields()
```
treykris, line 172: In order to assign a custom field value, we need to use the ID of that custom field and not the name. We don't have access to these custom field IDs unless we store it somewhere or request it. Maybe we can store them in JSON file instead of having this extra request

Here's an example response from the sendgrid api when we have a GET request for `https://api.sendgrid.com/v3/marketing/field_definitions`
```
[
      {
        id: 'e3_T',
        name: 'isBanned',
        field_type: 'Text',
        _metadata: {
          self: 'https://api.sendgrid.com/v3/marketing/field_definitions/e3_T'
        }
      }
    ]
```

This doesn't work in the PUT request in `putContact`
```
custom_fields: {
   isBanned: user.isBanned
}
```

but this does
```
custom_fields: {
   e3_T: user.isBanned
}
```

config.example.ts
```diff
@@ -25,6 +25,10 @@ const config: Static<typeof Config> = {
     unsubscribeGroup: {
       newsletter: 12567,
       account: 12570
+    },
+    contactList: {
+      students: '69f3f1c3-18e9-4ce0-b552-c28d8bd7bcee',
+      volunteers: '84a0bb98-a153-4897-9379-33016dccf921'
```
treykris, line 31: These two contact lists currently link to test contact lists - there are separate contact list IDs for prod

treykris: I think I might remove the test contact lists here too so that we don't keep adding contacts to a test list. We also have a 5k limit per month on the number of contacts that we can add to SendGrid

PR #430, opened 2020-08-06
Admin update user

treykris:
Links
-----
- Web repo PR: https://github.com/UPchieve/web/pull/497

Description
-----------
- Allow admins to update a user
- Update sendgrid when updating a user
- Format the partnerOrg data to the client with a key and displayName property

### DIFF
--- a/services/UserService.js
+++ b/services/UserService.js
@@ -2,6 +2,7 @@ const crypto = require('crypto')
 const { omit } = require('lodash')
 const User = require('../models/User')
 const Volunteer = require('../models/Volunteer')
+const Student = require('../models/Student')
 const MailService = require('./MailService')
 const UserActionCtrl = require('../controllers/UserActionCtrl')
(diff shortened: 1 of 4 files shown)


services/UserService.js
```diff
@@ -243,5 +244,62 @@ module.exports = {
...
+
+      if (partnerSite) update.partnerSite = partnerSite
+      else update.$unset.partnerSite = ''
+    }
+
+    // Remove $unset property if it has no properties to remove
+    if (Object.keys(update.$unset).length === 0) delete update.$unset
+
+    const updatedUser = Object.assign(userBeforeUpdate, update)
+    MailService.createContact(updatedUser)
```
treykris, line 297: I decided to update the user here using `Object.assign` since `Volunteer.updateOne` won't return a user below, but alternatively, I could probably use `Volunteer.findByIdAndUpdate` first and then pass the returned user to the MailService.createContact

jkeat: Object.assign seems good to me :+1:

PR #459, opened 2020-09-21
Add socket.io-redis

jkeat:
Description
-----------
- Add socket.io-redis

todo:
- refactor to remove `userSockets` from socketService


thedevelopnik:
@treykris Jack left this in draft mode, but looking at the [docs for this](https://github.com/socketio/socket.io-redis) I don't know how much more there is to do other than testing/validation in staging.

thedevelopnik (approved):

thedevelopnik:
We need to maintain a map of userIds to sockets, that is accessible across instances.

services/SocketService.js
```diff
@@ -35,10 +36,15 @@ module.exports = function(io) {
...
 
+      let userSockets = JSON.parse(await Redis.redisGet(userSocketsKey))
+      if (userSockets === null) {
+        userSockets = {}
+      }
       if (!userSockets[userId]) {
         userSockets[userId] = []
       }
       userSockets[userId].push(socket)
+      await Redis.redisSet(userSocketsKey, JSON.stringify(userSockets))
```
treykris, line 47: I’m getting a `TypeError` thrown when trying to stringify a user’s associated sockets:
`TypeError: Converting circular structure to JSON`

Looks like stringifying the sockets may be an issue and might have to find an alternative method to store the sockets in Redis

treykris:
I don't want to get ahead of myself, but it just hit me that this may be more work than anticipated. I believe we have a few different options, two of which are roughly similar. I would love to get your feedback on these approaches @jkeat, @thedevelopnik. Here's an overview of what I'm thinking:

1. We find a way to serialize the user sockets into Redis, keeping our approach with how we're using sockets the same. Currently, when trying to serialize a user's sockets into Redis we are getting errors related to circular references.
2. We store a user’s ID and the session room (the `sessionRoom` can be something like `sessions-${sessionId}`) that they belong to in Redis like so: `userID: sessionRoom`. When a user makes the initial connection to the socket server, we check Redis and if they have a key there, we will join them to that session room that is stored. When a user joins or starts a session, we set their `userId` to the `sessionRoom` they are joining. 
3. We store the `sessionRoom` as a key and the user’s that belong to room in Redis like so: `sessionRoom: [studentId, volunteerId]` (or maybe just some string denoting both user ids). When a user makes the initial connection to the socket server, we check that user's latest session, see if it's still active and connect them to the sessionRoom

- I think with option 1 and the way things are currently structured, we are manually doing things that Socket.io provides. We’re looping through all of a user's sockets and emitting from those sockets. I think if we utilized [rooms](https://socket.io/docs/rooms/) that should handle all of that for us. After some light digging around, I'm not too certain that storing a user's sockets in Redis is the direction to go anymore
- Option 2 and 3 will need a refactor on how we have our web sockets structured, but I think the refactor will lead to a simpler solution. In the refactor: User joins a room -> events are emitted to all that are subscribed to that room. We don’t need to worry about storing their sockets

treykris (approved):
Looks good!

Issue #473, opened 2020-09-30
Update remaining services to use TS


thedevelopnik:
Several services are still in JS, and need to be updated to use TS.

Posting mainly for tracking purposes. Probably should do one PR per controller to keep the effort moving and avoid huge PRs that are hard to audit, and use this issue to keep track of done-ness.

[ ] AWS
[ ] IneligibleStudent
[ ] IpAddress
[ ] Mail
[ ] Notification
[ ] PushToken
[ ] Queue
[ ] School
[ ] Session
[ ] Socket
[ ] Twilio
[ ] User
[ ] Whiteboard

PR #476, opened 2020-10-01
Converts PushToken service to TypeScript

thedevelopnik:
Links
-----
- Issue: https://github.com/UPchieve/server/issues/473

Description
-----------
- Converts the PushToken service to TS. Two main things that need review:
  - I've made an interface to represent a Session mongoose document at line 19, to be used on line 66 in the `sendVolunteerJoined` function. This is temporary until the Session model is converted to TS, then it could be brought in directly as a type. This whole PR might just want to wait until the model conversion PR is merged.
  - Converting actually surfaced a bad type in the android message object, where `priority` (line 49) is supposed to be a string of `high | normal` but was `1`. I guessed that `1` was supposed to imply `high` but if it was supposed to be equivalent to `normal` I can correct that.

Otherwise I think this is pretty close to done.

### DIFF
--- a/services/PushTokenService.ts
+++ b/services/PushTokenService.ts
@@ -1,7 +1,29 @@
-const admin = require('firebase-admin')
+import { messaging } from 'firebase-admin'
+import { Types } from 'mongoose'
 
-const sendToUser = ({ title, text, data, tokens }) => {
-  return admin.messaging().sendMulticast({
+// This interface feels cleaner than inlining it in the
+// sendToUser definition
+interface sendToUserData {
+  title: string
+  text: string
+  data: {
+    path: string
+  }
+  tokens: string[]
+}
+
+// Until the Session model is converted to TS,
+// we need to represent the parts of it used in
+// sendToVolunteer() here. When that model conversion
+// is done, this can be refactored to use that model directly.
+interface Session {
+  type: string
+  subTopic: string
+  _id: Types.ObjectId
+}
+
+const sendToUser = ({ title, text, data, tokens }: sendToUserData) => {
+  return messaging().sendMulticast({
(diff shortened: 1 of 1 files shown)


treykris:
Yeah, I agree I think the priority should be made as `high` - I took a look at the Firebase documentation and it wasn't as explicit to what the default is, but a priority of `high` is definitely the behavior we'd like. Thank you!

services/PushTokenService.ts
```diff
@@ -37,8 +62,8 @@ const sendToUser = ({ title, text, data, tokens }) => {
   })
 }
 
-module.exports = {
-  sendVolunteerJoined: async function(session, tokens) {
+export default {
```
treykris, line 65: We’ve come across this issue when importing functions from a TS file into a JS file where what's being exported actually gets wrapped in a `default` object as the namespace. In our TypeScript files, we've also needed to do a  `module.exports` alongside `export default` if they have functions being imported from a JS file

This issue came up on the recent contribution to `QueueService.ts` an error was thrown when the queue started up saying `queue.add is not a function` and it turned out the [twilioService](https://github.com/UPchieve/server/blob/85a472f32eb00631bbd837394c6a85ee2ad5affd/services/twilio.js#L6) was importing `queue` as:
```
{ default: queue }
```

I did a revert of QueueService back to JS on prod, but after digging into it a little bit I noticed what I mentioned above and made a PR resolving it in the meantime that's pending to be merged:
https://github.com/UPchieve/server/blob/2509105767228fb734c93687be3db00a19ca78b2/services/QueueService.ts#L6

https://github.com/UPchieve/server/pull/479

treykris (changes requested):

PR #479, opened 2020-10-01
Add cjs export ts

treykris:
Description
-----------
- Added `module.exports` to `QueueService.ts` for when importing from a JS file. In `services/twilio.js`, `queue` would be wrapped in a `default` object and `queue.add` would throw a TypeError of `TypeError: queue.add is not a function`: https://github.com/UPchieve/server/blob/85a472f32eb00631bbd837394c6a85ee2ad5affd/services/twilio.js#L357
- Lint


treykris:
Looks good to me!

PR #481, opened 2020-10-02
account deactivated user action

treykris:
Description
-----------
- User action for account deactivation
- Added new action `admin` to differentiate when a user deactivates their own account vs when an admin does


dmnisson (approved):
Looks good to me!

PR #489, opened 2020-10-15
Speed session queries

treykris:
Description
-----------
- Moved checking for the current session after checking conditionals 
- Updated the `getStaleSessions` query to only search for stale sessions in the window of time since its last run. Currently, it checks all sessions with a `createdAt` longer than 12 hours ago
- Have `getUnfullfilledSessions` utilize the compound index of `volunteer: 1, endedAt: 1` instead of the `createdAt: 1` index. This resulted in fewer documents and index keys examined, but it now sorts the sessions in memory
- Removed `$and` operator from `this.findLatest` query since mongoose find queries are an implicit `and`

### DIFF
--- a/controllers/SessionCtrl.js
+++ b/controllers/SessionCtrl.js
@@ -12,17 +12,15 @@ module.exports = function(socketService) {
       const userId = user._id
       const type = options.type
       const subTopic = options.subTopic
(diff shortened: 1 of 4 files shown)


treykris (commented):
Looks good!

PR #494, opened 2020-10-15
Converts IneligibleStudent model to typescript

thedevelopnik:
Links
-----
- Issue: https://github.com/UPchieve/server/issues/490

Description
-----------
- Converts IneligibleStudent model to typescript, from the large models PR


models/IneligibleStudent.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const ineligibleStudentSchema = new Schema({
+  createdAt: { type: Date, default: Date.now },
+  email: String,
+  zipCode: String,
+  school: {
+    type: Types.ObjectId,
+    ref: 'School'
+  },
+  ipAddress: String
```
treykris, line 23: Missing the `referredBy` property in the Schema and in the interface

treykris (changes requested):

models/IneligibleStudent.ts
thedevelopnik: Good catch, this is the other reason to do this one at a time, catching the little things that have gone out of date

thedevelopnik:
This now touches the User.ts model, but just enough to get this one to work. Will need to swing back around and make sure that's fully updated but didn't want to intertwine it to much with this now.

models/IneligibleStudent.ts
```diff
@@ -0,0 +1,38 @@
...
+import { User } from './User'
+
+export interface IneligibleStudent {
+  _id: Types.ObjectId;
+  createdAt: Date;
+  email: string;
+  zipCode: string;
+  school: Types.ObjectId | School;
+  ipAddress: string;
+  referredBy: Types.ObjectId | User
```
treykris, line 12: I think we might have some luck with this one! We will actually be able to get away without having the union type here. I couldn't find any place in the codebase that we're currently using the `referredBy` on this interface as a `User` type. I think we can hold off on it and leave a @todo comment for adding the union type with `User` and have it as:

```
referredBy: Types.ObjectId
```

 for now, until we get `User.ts` merged. Especially since it looks like `User.ts`  is also importing `import { Session } from './Session'` and then we'd have to also include those models `Session.ts` depends on

models/User.ts
```diff
@@ -3,6 +3,35 @@ import bcrypt from 'bcrypt';
 import validator from 'validator';
 import config from '../config';
 import { USER_BAN_REASON } from '../constants';
+import { IpAddress } from './IpAddress'
+import { Session } from './Session'
```
treykris, line 7: Now that I think about it - another option we can do is remove `pastSessions` on the user interface to avoid having to import `Session` here

models/IneligibleStudent.ts
```diff
@@ -0,0 +1,38 @@
+import { Document, model, Schema, Types } from 'mongoose';
+import { School } from './School';
+import { User } from './User'
```
treykris, line 3: Sorry for going comment crazy haha

Let's import this from `models/types` and put a comment to change it to import from User.ts when converted

thedevelopnik: I like that plan, I'll make the changes.

treykris (changes requested):
Looks good!!

treykris (approved):

PR #498, opened 2020-10-16
track cumulative stats

austinhallock:
added a couple of metrics to graph cumulative # of students and sessions


treykris (approved):
Looks good to me

PR #525, opened 2020-12-08
Fix elapsed availability

treykris:
Description
-----------
- Only calculate elapsed availability for users that are approved and onboarded
- Updated calculateElapsedAvailability tests
- DB util to set elapsedAvailability to 0 if not onboarded or approved


dbutils/fix-elapsed-availability.ts
```diff
@@ -0,0 +1,30 @@
...
+import VolunteerModel from '../models/Volunteer';
+
+const main = async (): Promise<void> => {
+  try {
+    await dbconnect();
+
+    const result = await VolunteerModel.updateMany(
+      {
+        isApproved: false,
+        isOnboarded: true,
```
treykris, line 12: Volunteers who are approved, but not onboarded should not have `elapsedAvailability` greater than 0. We will start calculating elapsedAvailability only until a volunteer has been onboarded and approved. There are edge cases to this with certain older volunteer accounts but we will ignore them for now and have them keep their current `elapsedAvailability`

thedevelopnik (approved):

PR #530, opened 2020-12-17
converts push token service to typescript

thedevelopnik:
Links
-----
- Issue: https://github.com/UPchieve/server/issues/473

Description
-----------
- Converts the PushToken service to TS. Two main things that need review:
  - I've made an interface to represent a Session mongoose document at line 19, to be used on line 66 in the `sendVolunteerJoined` function. This is temporary until the Session model is converted to TS, then it could be brought in directly as a type. This whole PR might just want to wait until the model conversion PR is merged.
  - Converting actually surfaced a bad type in the android message object, where `priority` (line 49) is supposed to be a string of `high | normal` but was `1`. I guessed that `1` was supposed to imply `high` but if it was supposed to be equivalent to `normal` I can correct that.

Otherwise I think this is pretty close to done.

### DIFF
--- a/services/PushTokenService.ts
+++ b/services/PushTokenService.ts
@@ -1,7 +1,29 @@
-const admin = require('firebase-admin')
+import { messaging } from 'firebase-admin'
+import { Types } from 'mongoose'
 
-const sendToUser = ({ title, text, data, tokens }) => {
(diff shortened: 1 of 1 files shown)


treykris (approved):
Looks good!

PR #533, opened 2020-12-18
Weekly summary emails

treykris:
Description
-----------
- Added a new cron job for weekly summary emails
- Added `VolunteerService` to aggregate needed data for the weekly summary emails
- Tests for `UserActionService`, `VolunteerService`, and `SessionService`

### DIFF
--- a/services/MailService.js
+++ b/services/MailService.js
@@ -339,6 +339,9 @@ module.exports = {
     )
   },
 
+  // @todo
+  sendWeeklySummaryEmail: data => {},
+
   createContact: async user => {
     const customFields = {
       [SG_CUSTOM_FIELDS.isBanned]: String(user.isBanned),
--- a/services/VolunteerService.ts
+++ b/services/VolunteerService.ts
@@ -1,0 +1,45 @@
+import VolunteerModel from '../models/Volunteer';
+import { getTimeTutoredForDateRange } from './SessionService';
+import { getElapsedAvailabilityForDateRange } from './AvailabilityService';
+import { getUnlockedSubjectsForDateRange } from './UserActionService';
+
+export const getVolunteers = async (
+  query,
+  projection = {}
+  // @todo: Use Volunteer interface once converted inside /models/Volunteer
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+): Promise<any[]> =>
+  VolunteerModel.find(query)
+    .select(projection)
+    .lean()
+    .exec();
+
+export const getWeeklySummaryStats = async (
+  volunteerId,
+  fromDate,
+  toDate
+): Promise<{
+  totalTimeTutored: number;
+  totalUnlockedSubjects: number;
+  totalElapsedAvailability: number;
+}> => {
+  // @todo: promise.all fails fast, do we want this? - handle error
+  const [
+    unlockedSubjects,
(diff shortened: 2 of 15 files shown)


worker/jobs/emailWeeklyHourSummary.ts
```diff
@@ -0,0 +1,75 @@
...
+    },
+    {
+      firstname: 1,
+      email: 1,
+      sentHourSummaryIntroEmail: 1
+    }
+  );
+
+  //  Monday-Sunday
+  const lastMonday = moment()
```
ti2ger92, line 27: Could variables be typed? If so, I think `lastMonday` could be `lastMonday: Moment`.

services/VolunteerService.ts
```diff
@@ -0,0 +1,50 @@
...
+  projection = {}
+  // @todo: Use Volunteer interface once converted inside /models/Volunteer
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+): Promise<any[]> =>
+  VolunteerModel.find(query)
+    .select(projection)
+    .lean()
+    .exec();
+
+export const getHourSummaryStats = async (
```
ti2ger92, line 17: Can we type inputs: `volunteerId: string`, etc...

worker/jobs/emailWeeklyHourSummary.ts
```diff
@@ -0,0 +1,75 @@
...
+      };
+      await MailService.sendHourSummaryEmail(data);
+      if (!sentHourSummaryIntroEmail)
+        await VolunteerModel.updateOne(
+          { _id },
+          { sentHourSummaryIntroEmail: true }
+        );
+      totalEmailed++;
+    } catch (error) {
+      log(
```
ti2ger92, line 68: This should probably roll up to console.error, so it can be alarmed on better, yes? To your logger, you could add a new function: error(message: string) which calls `console.error`.

Optionally, you may prefer Instead of a try / catch inside a for loop, use `await Promise.allSettled(), and with rejections, log the response. This will parallelize the work (if that is something you want), and IMO it's more readable.

services/MailService.js
```diff
@@ -339,6 +356,68 @@ module.exports = {
...
+    email,
+    sentHourSummaryIntroEmail,
+    fromDate,
+    toDate,
+    totalCoachingHours,
+    totalElapsedAvailability,
+    totalQuizzesPassed,
+    totalVolunteerHours
+  }) => {
+    const formatCoachHours = moment()
```
ti2ger92, line 370: Could these be typed? (are they strings?) Perhaps: `formattedCoachHours`?

services/MailService.js
```diff
@@ -83,6 +86,17 @@ const buildLink = path => {
   return `${protocol}://${host}/${path}`
 }
 
+const getFormattedHourSummaryTime = ([hours, minutes]) => {
```
ti2ger92, line 89: If at all possible, I would get rid of this array, instead accept `time`, and format it entirely here. In Typescript, I'd always use an interface, instead of a 2-value array, but if you're already getting your values as a formatted string, I'd just leave all the formatting logic here.

ti2ger92 (changes requested):
I like all the use of Typescript.

